PR #2, opened 2018-12-15
[WIP] Full ecosystem setup based on full-stack example

ahmet:
Hi,

We're currently exploring ORY ecosystem in order to decide whether we should use or not at [my current company](https://github.com/biges).

During this experiments we wrote these Kubernetes configs based on docker-compose.yml of [full-stack example](https://github.com/ory/examples/blob/master/full-stack/docker-compose.yml) and we thought that sharing is caring 😌

It's still a work in progress though.

Keep up the good work 👏🏼


yaml/oathkeeper/simple/oathkeeper-api.yaml
```diff
@@ -1,51 +0,0 @@
----
```
aeneasr: The general idea of this repository to be a *collection* of different ways to solve things. In fact, if you remove this specific file, several published blog posts would fail. Maybe it would make more sense to have e.g. another directory like `yaml/full-stack/postgres/...` where the files you published reside in?

ahmet: Putting it all in directory named yaml doesn't make sense IMHO, as a shell script to generate secrets dynamically included for now and more would come in future.

How about reverting that file back and moving full-stack example related files under a dir named full-stack, just like [docker based examples](https://github.com/ory/examples/tree/master/full-stack)?

aeneasr: makes sense!

aeneasr:
Thank you for your hard work on this PR. It is however superseded by the helm charts we introduced. It may not set everything up 1:1 on par with the examples directory but it offers the ability to set it up. With helm 3.0 (and already now) we can use template rendering so Tiller is no longer required, thus making these charts as usable as raw kubernetes resource files!

PR #3, opened 2019-04-06
Helm chart for Hydra

ermik:
WIP — you can see current state of the chart in `.rendered.yaml` file.

- [ ] Implement Kubernetes scalability controls
- [ ] Add Helm charts for each of oryOS components, individually deployable as well as set as Helm sub-chart dependencies of the main Hydra chart.
- [ ] Adopt new Hydra `yaml` configuration as a `ConfigMap` file embedding — e.g.
  ```yaml
  apiVersion: v1
  kind: ConfigMap
  data:
    ".hydra.yaml": |-
      serve:
        public:
          port: 4444
      ...
  ```
  or use `{{ tpl }}` function to pre-generate yaml resource file:
  ```yaml
  apiVersion: v1
  kind: ConfigMap
  data:
    ".hydra.yaml": |-
       {{- tpl (.Files.Get "resources/.hydra.yaml.tpl") . | nindent 4 }}
  ```
- [ ] Find a way to run `helm template` command generating `.rendered.yaml` snapshot and commit it from the CI

as per request in Discord on 03/30/2019

Non-goals (will be accomplished in the follow-up PRs):
- usage of Helm test utilities
- dependency on and compatibility with non-Ory charts (databases, ambassador proxy, istio, etc.)
- compatibility with global Chart repo — after we're done with v0.1.0 we can push it there to start getting feedback of the community at-large

### DIFF
--- a/helm/hydra/.rendered.yaml
+++ b/helm/hydra/.rendered.yaml
@@ -72,7 +72,7 @@ spec:
             - name: SYSTEM_SECRET
               valueFrom:
                 secretKeyRef:
-                  name: hydra-secret
+                  name: hydra-system-secret
                   key: system
       containers:
         - name: "hydra-public"
@@ -109,7 +109,7 @@ spec:
             - name: SYSTEM_SECRET
               valueFrom:
                 secretKeyRef:
-                  name: hydra-secret
+                  name: hydra-system-secret
                   key: system
 
         - name: "hydra-admin"
@@ -146,7 +146,7 @@ spec:
             - name: SYSTEM_SECRET
               valueFrom:
                 secretKeyRef:
-                  name: hydra-secret
+                  name: hydra-system-secret
                   key: system#
     #
     #
--- a/helm/hydra/.rendered.yaml
+++ b/helm/hydra/.rendered.yaml
@@ -1,0 +1,153 @@
+---
+# Source: hydra/templates/service.yaml
+apiVersion: v1
+kind: Service
+metadata:
+  labels:
+    app.kubernetes.io/name: hydra
+    helm.sh/chart: hydra-0.0.1
+    app.kubernetes.io/managed-by: Tiller
+    app.kubernetes.io/instance: hydra
+  name: hydra
+spec:
+# Provides options for the service so chart users have the full choice
+  type: "ClusterIP"
+  ports:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 4444
+  selector:
+    app.kubernetes.io/name: hydra
+    app.kubernetes.io/instance: hydra
+
+---
+# Source: hydra/templates/deployment.yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: hydra
+  labels:
+    app.kubernetes.io/name: hydra
+    helm.sh/chart: hydra-0.0.1
+    app.kubernetes.io/version: 0.0.1
+    app.kubernetes.io/instance: hydra
+    app.kubernetes.io/managed-by: Tiller
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: hydra
+      app.kubernetes.io/instance: hydra
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/name: hydra
+        app.kubernetes.io/instance: hydra
+    spec:
+      initContainers:
+        - name: hydra-migrate
+          image: "oryd/hydra:v1.0.0-rc.8_oryOS.10"
+          imagePullPolicy: IfNotPresent
+          command: ["hydra"]
+          args: [
+              "migrate",
+              "sql",
+              "postgres://$(DB_USERNAME):$(DB_PASSWORD)@$(DB_HOST):$(DB_PORT)/$(DB_NAME)"
+            ]
+          envFrom:
+            - configMapRef:
+                name: "hydra-env"
+          env:
+            - name: DB_USERNAME
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-db-secrets
+                  key: username
+            - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-db-secrets
+                  key: password
+            - name: SYSTEM_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-secret
+                  key: system
+      containers:
+        - name: "hydra-public"
+          image: "oryd/hydra:v1.0.0-rc.8_oryOS.10"
+          imagePullPolicy: IfNotPresent
+          command: ["hydra"]
+          args: ["serve", "public"]
+          livenessProbe:
+            httpGet:
+              path: "/health/alive"
+              port: "http"
+          readinessProbe:
+            httpGet:
+              path: "/health/ready"
+              port: "http"
+          ports:
+            - name: "http-public"
+              containerPort: 4444
+              protocol: TCP
+          envFrom:
+            - configMapRef:
+                name: "hydra-env"
+          env:
+            - name: DB_USERNAME
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-db-secrets
+                  key: username
+            - name: DB_PASSWORD
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-db-secrets
+                  key: password
+            - name: SYSTEM_SECRET
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-secret
+                  key: system
+
+        - name: "hydra-admin"
+          image: "oryd/hydra:v1.0.0-rc.8_oryOS.10"
+          imagePullPolicy: IfNotPresent
+          command: ["hydra"]
+          args: ["serve", "admin"]
+          livenessProbe:
+            httpGet:
+              path: "/health/alive"
+              port: "http"
+          readinessProbe:
+            httpGet:
+              path: "/health/ready"
+              port: "http"
+          ports:
+            - name: "http-admin"
+              containerPort: 4445
+              protocol: TCP
+          envFrom:
+            - configMapRef:
+                name: "hydra-env"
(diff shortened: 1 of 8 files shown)


ermik:
@aeneasr I couldn't reach you on Discord — I've had Hydra running in kubernetes for half a year or so now, and I wanted to convert the manifests I use to a Helm chart for a while now.

aeneasr:
Hey, this is great! I have zero knowledge of helm so what should I look out for when reviewing?

helm/hydra/.rendered.yaml
```diff
@@ -0,0 +1,153 @@
...
+    spec:
+      initContainers:
+        - name: hydra-migrate
+          image: "oryd/hydra:v1.0.0-rc.8_oryOS.10"
+          imagePullPolicy: IfNotPresent
+          command: ["hydra"]
+          args: [
+              "migrate",
+              "sql",
+              "postgres://$(DB_USERNAME):$(DB_PASSWORD)@$(DB_HOST):$(DB_PORT)/$(DB_NAME)"
```
aeneasr: Is this linked to postgres? Also keep in mind that we can use `-e` to load this from the enfironment ($DSN)

helm/hydra/.rendered.yaml
```diff
@@ -0,0 +1,153 @@
...
+              valueFrom:
+                secretKeyRef:
+                  name: hydra-secret
+                  key: system
+      containers:
+        - name: "hydra-public"
+          image: "oryd/hydra:v1.0.0-rc.8_oryOS.10"
+          imagePullPolicy: IfNotPresent
+          command: ["hydra"]
+          args: ["serve", "public"]
```
aeneasr: Is there an upside to running those two commands separately? Keep in mind that it won't work with DSN=memory if split.

ermik:
@aeneasr So, briefly, Helm could be called a cluster management procedure — cluster administrator uses an abstraction of `values.yaml` file (or `helm` CLI args to override) at to specify the settings for a given resource and then deploy it in a controlled manner (all resources stay bundled in the cluster via usage of `helm.sh` annotations, can be rolled back, version-controlled, upgraded). 

User the "hood" Helm utility workflow uses `go-template` functionality to create said manifest from a set of template files and then uses the `Tiller` Pod (deployed to a cluster) to orchestrate the deployment or upgrade.

First and foremost our priority should be providing a well-structured API (of sorts) in the `values.yaml` file — so that end-user can easily tweak the deployment to suit their individual needs without resorting to editing the template files.

When reviewing, it's easiest to look at `.rendered.yaml` which is a snapshot of manifests Helm will end up committing to the Kubernetes cluster and `values.yaml` as a snapshot of *functionality* we provide to users of this Helm chart. 

Also, keep in mind that you are free to edit `values.yaml` as part of this PR — if you use a higher-level (author's) perspective to provide the _wishlist_ of settings there, I will then follow-through by implementing the template-level instructions to make those settings actually change the result.

helm/hydra/.rendered.yaml
ermik: At the time of this initial commit the source Chart (private to my org) uses Hydra at rev pre-dating the `.hydra.yaml` and constrained to PostgreSQL. _It is my plan to remove this constraint within the scope of this PR._

When it comes to envvars we have to be careful to walk the line between convenience and readability. From my perspective: for a cluster administrator unfamiliar with Hydra, but still required to provide the SQL storage `Resource` for whoever is requesting Hydra to be deployed, it will be better to 
- (1) explicitly see the source of DSN and 
- (2) to have a single-glance of the environment that has a major impact on the ability of Hydra to run
One way we achieve that is my opting in to listing critical environment variables directly in `PodSpec` (`deployment.yaml`) rather than _outsourcing_ them to `hydra-env` `ConfigMap`. Perhaps having an appearance of:
```
ENV1="env1"
ENV2="env2"
...
DSN="$(ENV1) plus $(ENV2) and ..."
```
will be [better][rabbitmq-example] than setting it via `args` — it's up to us to decide.

[rabbitmq-example]: https://github.com/helm/charts/blob/7b1290efb7172de960b43a9f8bc603e4c38e2594/stable/rabbitmq/templates/statefulset.yaml#L159-L179

helm/hydra/.rendered.yaml
ermik: This is the heritage of how we run Hydra — we have 3 _sets_ of containers running:
1. `hydra-admin` — 2 or 3 copies ran as `serve admin` for minimum fault-tolerance, only available inside the cluster
2. `hydra-intern` — N copies of `serve public` scaled to the needs of internal clients, only available inside the cluster
3. `hydra-public` — M copies of `serve public` scaled to the needs of the public clients, available on the edge to anyone who hits the `OAUTH2_ISSUER_URL`

Initial commit to this PR does not implement the scalability per-se, but it showcases our ability to provide Helm chart users with an option:
https://github.com/ory/k8s/blob/33d2d8b0681fd2f845b16a8d3a1d21069be5db66/helm/hydra/values.yaml#L23
which will result in manifest output you see `.rednered.yaml` if set to `true`, and if set to `false` will generate `PodSpec` with a *single container* running `hydra serve both`.

All of this is TBD.

ermik:
I use this article during chart development a lot (for reference):
https://github.com/helm/helm/blob/master/docs/charts_tips_and_tricks.md

This article describes values file usage and ways to override:
https://github.com/helm/helm/blob/master/docs/chart_template_guide/values_files.md

helm/hydra/templates/_helpers.tpl
```diff
@@ -71,7 +71,7 @@ readinessProbe:
 - name: SYSTEM_SECRET
   valueFrom:
     secretKeyRef:
-      name: hydra-secret
+      name: {{ required "Hydra Helm chart requires `systemSecretName` to be specifed and persisted in Kubernetes. See README for more information" .Values.systemSecretName }}
```
ermik: The `values.yaml` key wasn't used yet previously; now it is. Example of the results (I'm using `helm template` but it could be `helm install`):

Error:
```
helm template \
    --name hydra \
    --namespace hydra ./ > .rendered.yaml

Error: render error in "hydra/templates/deployment.yaml": template: hydra/templates/deployment.yaml:39:14:
  executing "hydra/templates/deployment.yaml" at <include "hydra.secre...>: 
    error calling include: template: hydra/templates/_helpers.tpl:74:15: 
      executing "hydra.secrets" at <required "Hydra requ...>: 
        error calling required: Hydra Helm chart requires `systemSecretName` to be specifed and persisted in Kubernetes. See README for more information.

(this output was wrapped for display on GitHub; original line count: 2)
```

OK:
```
helm template \
  --name hydra \
  --namespace hydra \
  --set systemSecretName=hydra-system-secret ./ > .rendered.yaml
```

helm/hydra/.rendered.yaml
aeneasr: I think that's a great idea with the singleton - we should probably default to `true`.

helm/hydra/templates/_helpers.tpl
aeneasr: Awesome!

aeneasr:
So I thought a bit about what we could do with helm. There are a few core aspects we need to take care of:

- `secrets.*` - These values must be set using k8s secrets and should be referenced in a way that is already being done [here](https://github.com/ory/k8s/pull/3#discussion_r279201747)
- `urls.*` - These values must be set using a configuration setting. They are unique to each deployment unless we provide, for example, [demos / fallbacks](#urls-defaults).
- `serve.<admin|public>.cors` should be configurable with [sane defaults](#cors-sane-defaults).
- `serve.tls` I think that most deployments will not make use of Hydra's integrated TLS functionality. Instead they will set `allow_termination_from` and terminate TLS at the edge. This should somehow be easily configurable.
- `tracing` we might want to set up the ability to run traces without a lot of work. Not sure how :)

On top of that we might want to think about JWT Strategies. However, it is very bad practice to use JWTs for access tokens. Instead, this might be an area where we would add Oathkeeper. But let's work on the Hydra helm chart first and just always use the opaque strategy.

I will think about more ways to integrate keto/oathkeeper in the future here as well. Just need some deep thinking :)

## URLS Defaults

I think we could deploy the exemplary login/consent/logout app alongside ORY Hydra in case where it's not explicitly specified. It should probably be something very visibly like `demo-integration: true/false`.

## CORS Sane Defaults

The idea would be to have configuration elements in values.yml like `cors_enable_public`, `cors_enable_admin` (naming up to you). Additionally, there needs to be `allowed_origins` set (`serve.<admin|public>.cors.allowed_origins`) to the origins where CORS is allowed from. All other settings should be set with defaults, if enabled:

```
      allowed_headers:
        - Authorization
      exposed_headers:
        - Set-Cookie
        - Content-Type
      allow_credentials: true
      allowed_methods:
        - POST
        - GET
        - PUT
        - PATCH
        - DELETE
```

It might make more sense to set these defaults *in ORY Hydra* itself, as per issue https://github.com/ory/hydra/issues/1400

ermik:
upd: Slowly working out the kinks. Just a spare time being scarce, no other issues.

jayme-github:
There are two quite similar projects going on at https://github.com/helm/charts/pull/12845 and https://github.com/helm/charts/pull/11703
IMHO we should try to coordinate the efforts a bit. Any thoughts on that?

ermik:
I don't see much similarity as to the project state, but I do agree there is a lot of efforts out there. I have nothing against collaborating on the tasks outlined above.

aeneasr:
This has been bumped in priority on our side, and I will step in next week. Would it be possible for you to push your most recent changes? I would then branch off of this and continue your hard and awesome work :)

ermik:
Feel free to close, have been juggling too much non-open-source to really make an impact. Do reach out if you run into any issues, as I definitely have plenty of time to consult.

ermik:
I see that no code made it into Ory's take on Helm, so I will clear the source branch.

PR #8, opened 2019-06-17
helm: ORY Oathkeeper chart

aeneasr:


helm/ory-oathkeeper/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: v1
+kind: Pod
+metadata:
+  name: "{{ include "ory-oathkeeper.fullname" . }}-test-connection"
+  labels:
```
piotrmsc: Question: you have liveness and readiness probes defined on a deployment already so there is no need to duplicate it in the test.   Service definition is optional for `api` and `proxy` and the test here is making wget over `FQDN` which requires service to be created, so if the user will disable for example `api` this test will fail. What do you think?

helm/ory-oathkeeper/README.md
```diff
@@ -0,0 +1,84 @@
...
+### Access Rules
+
+Instead of fetching access rules from remote locations, you can set your access rules directly with `--set-file`:
+
+```bash
+$ helm install \
+    --set-file oathkeeper.accessRules=./path/to/access-rules.json \
+    ory-oathkeeper
+```
+
```
piotrmsc: Would be also nice to havet `NOTES` file which is rendered after installing the release via helm and includes information like how service can be accessed or managed.

helm/ory-oathkeeper/templates/deployment.yaml
```diff
@@ -0,0 +1,79 @@
...
+          env:
+            {{- if .Values.oathkeeper.mutatorIdTokenJWKs }}
+            - name: MUTATORS_ID_TOKEN_JWKS_URL
+              value: "file:///etc/secrets/mutator.id_token.jwks.json"
+            {{- end }}
+            {{- if .Values.oathkeeper.accessRules }}
+            - name: ACCESS_RULES_REPOSITORIES
+              value: "file:///etc/config/access-rules.json"
+            {{- end }}
+          volumeMounts:
```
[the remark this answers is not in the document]
aeneasr: I think (?) that's how it works. If you set `Values.oathkeeper.config.access_rules.repositories=http://remote/json` it should work. However, if you provide it via `oathkeeper.accessRules` directly, the above value will not work as it will be overwritten. I tried to document that in the README

helm/ory-oathkeeper/templates/tests/test-connection.yaml
```diff
@@ -0,0 +1,19 @@
...
+{{ include "ory-oathkeeper.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": test-success
+spec:
+  containers:
+    - name: liveliness-probe
```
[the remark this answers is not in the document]
aeneasr: Ah yeah :)

helm/ory-oathkeeper/templates/tests/test-connection.yaml
aeneasr: Hm yeah that's true. I wanted to have some type of programmatic confirmation using `helm test` to see if the deployment is working. What other options do we have here? I thought about maybe fetching `/rules` or `/.well-known/jwks.json`?

helm/ory-oathkeeper/README.md
aeneasr: A notes file exists! :)

aeneasr: helm/ory-oathkeeper/templates/NOTES.txt

helm/ory-oathkeeper/templates/tests/test-connection.yaml
piotrmsc: So here can be a pod which is making wget/curl for rules or even e2e scenario you already have developed :)  which is even more awesome :D but I know e2e will require much more stuff here so maybe start with fetching rules & JWKS

helm/ory-oathkeeper/README.md
piotrmsc: :D you are right! I have collapsed everything and didn't notice, my bad :)

aeneasr: No worries! :)

helm/ory-oathkeeper/templates/secret-jwks.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
```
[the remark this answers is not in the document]
aeneasr: Yeah this was mostly out of laziness on my part. I didn't want to add the if's in the volume mounts everywhere, so I thought just having an empty secret is ok. What do you think?

helm/ory-oathkeeper/templates/tests/test-connection.yaml
aeneasr: I really wanted to test ingress and LoadBalancer but since Minikube doesn't support those I wasn't able to (I was not able to confirm that those two set ups actually work yet). In those cases however, we could probably just do a `curl https://ingress/` with a LoadBalancer port behind it?

Maybe I'll set up a k8s cluster for testing and add it to the CI...what ya think?

piotrmsc: Fetching JWKS seems enough then I think. Regarding loadbalancer on a minikube, in the kyma on a minikube we use `hostPort` 80&443 in ingress controller deployment. So if we have oathkeeper exposed on host `oathkeeper.kyma.local` and make a curl to it from host machine, ingress controller is able to route inside the cluster such request. But for sure this is sth that can be done in the future 👍

helm/ory-oathkeeper/templates/secret-jwks.yaml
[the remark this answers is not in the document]
piotrmsc: not a problem ;-) 👍

helm/ory-oathkeeper/templates/_helpers.tpl
```diff
@@ -0,0 +1,45 @@
...
+{{- define "ory-oathkeeper.name" -}}
+{{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
+{{- end -}}
+
+{{/*
+Create a default fully qualified app name.
+We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
+If release name contains chart name it will be used as a full name.
+*/}}
+{{- define "ory-oathkeeper.fullname" -}}
```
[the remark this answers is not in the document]
aeneasr: It was generated by `helm charts` like that so I guess we'll just leave it to change as few default as possible?

PR #27, opened 2019-07-22
helm/oathkeeper: add extraConfigFiles and extraEnv

hypnoglow:
## Proposed changes

This commit adds more flexibility in oathkeeper helm chart
configuration. Now any number of additional configs and environment
variables can be passed.

## Checklist

### DIFF
--- a/helm/charts/oathkeeper/values.yaml
+++ b/helm/charts/oathkeeper/values.yaml
@@ -111,3 +111,19 @@ tolerations: []
 
 # Configure node affinity
 affinity: {}
+
+# Extra config files that will be put to the volume at "/etc/config".
+# For example, you can define multiple additional access rules files.
+extraConfigFiles: []
+  # Example:
+  # - name: access-rules.yaml
+  #   contents: |
+  #     - id: management-api
+  #       authenticators:
+  #         - handler: noop
+
+# Extra environment variables that will be passed to the container.
+extraEnv: []
+  # Example:
+  # - name: ACCESS_RULES_REPOSITORIES
+  #   value: file:///etc/config/access-rules.yaml
--- a/docs/helm/oathkeeper.md
+++ b/docs/helm/oathkeeper.md
@@ -82,3 +82,20 @@ $ helm install \
 
 Please note that any configuration values set for `oathkeeper.config.access_rules.repositories` using e.g.
 a configuration file will be overwritten by this setting.
+
+### Additional configuration files and environment variables
+
+If you need more flexibility with configuration files (e.g. multiple access rules, access rules in YAML format, etc),
+you can use `extraConfigFiles` to pass additional configs via values-file or helm flags.
+
+You can also provide additional environment variables using `extraEnv`.
+
+Example:
+
+```bash
+$ helm install ory/oathkeeper \
+    --set "extraConfigFiles[0].name=more-access-rules.yaml" \
+    --set-file "extraConfigFiles[0].contents=additional_access_rules.yaml" \
+    --set "extraEnv[0].name=ACCESS_RULES_REPOSITORIES" \
+    --set "extraEnv[0].value=file:///etc/config/access-rules.json,file:///etc/config/more-access-rules.yaml"
+```
--- a/helm/charts/oathkeeper/Chart.yaml
+++ b/helm/charts/oathkeeper/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: "v0.16.0-beta.4_oryOS.12"
+appVersion: "v0.17.0-beta.1"
 description: A Helm chart for deployoing ORY Oathkeeper in Kubernetes
 name: oathkeeper
 version: 0.0.1
--- a/helm/charts/oathkeeper/templates/configmap.yaml
+++ b/helm/charts/oathkeeper/templates/configmap.yaml
@@ -59,4 +59,7 @@ data:
 {{- with .Values.oathkeeper.config -}}
   {{- toYaml . | nindent 4 }}
 {{- end -}}
+{{- range $item := .Values.extraConfigFiles }}
+  {{ $item.name }}: |- {{ $item.contents | nindent 4 }}
+{{- end }}
 {{- end -}}
--- a/helm/charts/oathkeeper/templates/deployment.yaml
+++ b/helm/charts/oathkeeper/templates/deployment.yaml
@@ -44,6 +44,9 @@ spec:
             - name: ACCESS_RULES_REPOSITORIES
               value: "file:///etc/config/access-rules.json"
             {{- end }}
+            {{- if .Values.extraEnv }}
+            {{- toYaml .Values.extraEnv | nindent 12 }}
+            {{- end }}
           volumeMounts:
             - name: {{ include "oathkeeper.name" . }}-config-volume
               mountPath: /etc/config


helm/charts/oathkeeper/values.yaml
```diff
@@ -111,3 +111,19 @@ tolerations: []
...
+extraConfigFiles: []
+  # Example:
+  # - name: access-rules.yaml
+  #   contents: |
+  #     - id: management-api
+  #       authenticators:
+  #         - handler: noop
+
+# Extra environment variables that will be passed to the container.
+extraEnv: []
```
aeneasr: All config variables are loaded from `oathkeeper.config`. Do we really need to have another way of defining config? This might in fact be a problem where you define a custom env var, forget about it, update a config key in `oathkeeper.config` and are wondering the heck why it doesn't update the config while in fact the old env var is overriding the value.

helm/charts/oathkeeper/values.yaml
```diff
@@ -111,3 +111,19 @@ tolerations: []
 
 # Configure node affinity
 affinity: {}
+
+# Extra config files that will be put to the volume at "/etc/config".
+# For example, you can define multiple additional access rules files.
+extraConfigFiles: []
```
aeneasr: Would it make more sense if we changed `accessRules` from `Rule (Object)` to:

```yaml
oathkeeper:
  accessRules:
    - file-a.yaml
    - file-b.yaml
```

?

hypnoglow: I like the idea. But this is not enough, because we need both file name and file contents. 

Option 1)

```yaml
oathkeeper:
  accessRules:
    rules.yaml: |-
      - id: foo
```

But in this case you need to be careful when using `--set-file` because periods must be escaped: `--set-file "oathkeeper.accessRules.foo\.json=./helm/charts/oathkeeper/demo/access-rules.json"`

Option 2)

```yaml
oathkeeper:
  accessRules:
    - filename: rules.yaml
      contents: |-
        -id: foo
```

This is more tedious but no extra escaping required.

---

Also, I suggest introducing another parameter for `values` file where you can write rules not as a string, but as yaml (recently I added yaml support for rules here: https://github.com/ory/oathkeeper/pull/213), for example:

```yaml
oathkeeper:
  accessRules:
    - id: foo
      authorizer:
        handler: noop
```

I see this is by far more convenient than yaml strings with `|-`.

So we can have 2 paramerers: one for inline yaml and another for files, e.g.

```yaml
oathkeeper:
  accessRules:
    - id: foo
      authorizer:
        handler: noop
  accessRulesFiles:
      rules.yaml: |-
        - id: foo
```

@aeneasr  What do you think?

helm/charts/oathkeeper/values.yaml
hypnoglow: If user defines everything via config this is fine. But sometimes envs are more convenient, personally, I configure hydra purely via env vars :)

> This might in fact be a problem where you define a custom env var, forget about it, update a config key in oathkeeper.config and are wondering the heck why it doesn't update the config while in fact the old env var is overriding the value.

Yeah, this is always a choice whether to limit configuration options for the user safety or to allow flexible configuration. By the way, as I noticed the trend, charts official helm charts repository tend to provide as much flexibility as possible. 

So we have to choose our way of configuration :)

aeneasr: Since `oathkeeper.config` is at the core of this chart and also seeing that it's possible to change these config values _without restarting oathkeeper_ (it now supports hot reloading) I think that we should avoid confusion and not include `extraEnv` where hot reloading does not work (changing anything requires a restart of oathkeeper). What do you think?

helm/charts/oathkeeper/values.yaml
aeneasr: > I like the idea. But this is not enough, because we need both file name and file contents.

Can we not extract the filename from above? E.g. with a hash?

```
oathkeeper:
  accessRules:
    - file-a.yaml
    - file-b.yaml
```

Results in contents from `file-a.yaml` with filename `md5("file-a.yml")` and `file-b.yaml` with filename `md5("file-b.yml")`. What do you think?

> Also, I suggest introducing another parameter for values file where you can write rules not as a string, but as yaml (recently I added yaml support for rules here:

I think this could be a nice addition later on (e.g. `accessRuleInline`) but I'd leave it out of this PR for now and see if there are use cases where people demand this :)

hypnoglow: > Can we not extract the filename from above? E.g. with a hash?

The main problem is how to get the file contents, not the file name. Helm is very limited in reading files by name reference, so we definitely should support `--set-file` instead.

aeneasr: > The main problem is how to get the file contents, not the file name. Helm is very limited in reading files by name reference, so we definitely should support --set-file instead.

Ohh I see - that makes absolute sense. How about Option 3 then:

```yaml
oathkeeper:
  accessRules:
    - rules: |-
      - id: rule-1
        authenticator: ...
      - id: rule-2
        authenticator: ...
```

where the filename is generated by (`md5(accessRules.[*].rules)`)?

hypnoglow: So this is Option 1 basically, not the 3rd one (still using stirngs via `|-`).

In this case, why do we need md5? `rules` word in your example can be a filename:

```yaml
oathkeeper:
  accessRules:
    - rules.yaml: |-
      - id: rule-1
        authenticator: ...
      - id: rule-2
        authenticator: ...
```

aeneasr: Good point, I was thinking that we might want to have other options too, such as:

```
oathkeeper:
  accessRules:
    - onlyForAnntation: 
       - annotation/foo # this access rule would only be deployed on pods with this annotation
       rules: |-
       ...
```

But that's maybe a stretch, it would however allow for easy changes later on. Adding such a value would be difficult in the

```
oathkeeper:
  accessRules:
    - rules.yaml: |-
      - id: rule-1
        authenticator: ...
      - id: rule-2
        authenticator: ...
```

model because we don't know what the file name is and what a config option is.

hypnoglow: Personally, I think an example with annotations adds unnecessary complexity (the same issue with `md5` stuff instead of explicit filename).

> Adding such a value would be difficult in the ... model because we don't know what the file name is and what a config option is.

I'm not sure I understand the difficulty, isn't it simply `--set-file "oathkeeper.accessRules[0].rules\.yaml=somefile.yaml" `?

---

Let's sum up what we have. I propose one of the following:

1. Option with "string" config formatting, one of two variants:

    1. First

        ```yaml
        oathkeeper:
        accessRules:
          rules.yaml: |-
            - id: foo
        ```

        How to configure:

        1. Directly in the values.yaml file

        1. Using `--set-file "oathkeeper.accessRules[0].rules\.yaml=somefile.yaml"`

        Examples in other charts:

        - [rabbitmq](https://github.com/helm/charts/blob/master/stable/rabbitmq-ha/values.yaml#L32)
        - [fluentd](https://github.com/helm/charts/blob/master/stable/fluentd/values.yaml#L76)

    1. Second

        ```yaml
        oathkeeper:
        accessRules:
          - filename: rules.yaml
            contents: |-
              -id: foo
        ```

        How to configure:

        1. Directly in the values.yaml file

        1. Using `--set "oathkeeper.accessRules[0].filename=rules.yaml"` and `--set-file "oathkeeper.accessRules[0].contents=somefile.yaml"`

        Examples:

        - [coredns](https://github.com/helm/charts/blob/master/stable/coredns/values.yaml#L120) zone files

2. Option with "inline" yaml config formatting, similar to how `oathkeeper.config` section:

    ```yaml
    oathkeeper:
    accessRules:
      - id: foo
        authorizer:
          handler: noop
    ```

    This approach used widely in helm charts when the application supports yaml (or compatible) configuration as it provides native configuration using both in values.yaml file and using `--set` arguments.

    Examples:

    - [prometheus](https://github.com/helm/charts/blob/master/stable/prometheus/values.yaml#L981)
    - [dex](https://github.com/helm/charts/blob/master/stable/dex/values.yaml#L134)
    - [grafana](https://github.com/helm/charts/blob/master/stable/grafana/values.yaml#L333)
    - [mongodb](https://github.com/helm/charts/blob/master/stable/mongodb/values.yaml#L332)
    - [influxdb](https://github.com/helm/charts/blob/master/stable/influxdb/values.yaml#L135)

So what should we stick into?

aeneasr: Sorry for my slow response, there was so much stuff going on in the last couple of days.

Thank you for the great sum up. In retrospective, I think all 3 approaches are great. I think we should go with the approach that is most common, so here's my final proposal.

We change `oathkeeper.accessRules` from a single file to an array of files with yaml:

```yaml
oathkeeper:
accessRules:
  rules.yaml:
    - id: foo
      authorizer:
        handler: noop
```

Does that sound good to you?

PR #34, opened 2019-07-29
Replace GKE with kind

aeneasr:


.circleci/config.yml
```diff
@@ -13,18 +13,24 @@ jobs:
...
-          eval "$KUBECTL_COMMAND"
-      - run: kubectl get pods
+      - setup_remote_docker
+      - run: curl -LO https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl
+      - run: chmod +x ./kubectl
+      - run: mv kubectl /go/bin/kubectl
       - run: curl -L https://git.io/get_helm.sh | bash
-      - run: |
-          helm init --service-account $TILLER_SERVICE_ACCOUNT --tiller-namespace $TILLER_NAMESPACE
+      - run: make kind-start
```
Demonsthere: `/bin/bash: kind: command not found` Seems `make kind-start` doesn't install the kind binary. Btw, i think we may run of our resource here, as kind will try to launch a whole k8s. Consider increasing the [resource class](https://circleci.com/docs/2.0/configuration-reference/#resource_class)

aeneasr:
#35

PR #37, opened 2019-08-01
Update ory charts

Demonsthere:
## Proposed changes

- Change master to oathkeeper-master subchart
- Fix some small errors in all charts
- Unify common values across all charts
- Change hydra secret behaviour:
  - If no secret is given, generate a random one
  - Move secret creation to a pre-install hook, so it won't be touched during upgrades
- Update docs for all charts

## Checklist


helm/charts/oathkeeper/charts/oathkeeper-master/templates/deployment.yaml
```diff
@@ -41,7 +39,7 @@ spec:
             - /manager
           args:
             - --metrics-addr=127.0.0.1:8080
-            - --rulesConfigmapName={{ .Values.configMapName }}
+            - --rulesConfigmapName={{ .Release.Name }}-rules
```
aeneasr: I think this needs to be `oathkeeper.fullname` right?

Demonsthere: no, as this is a subchart, it inherits the name of its parent, so `fullname` produces `oathkeeper-oathkeeper-maester`

Demonsthere: After some thought I changes this to a smarter function

helm/charts/hydra/templates/secrets.yaml
```diff
@@ -5,8 +5,13 @@ metadata:
...
+  annotations:
+    # Create the secret before installation, and only then. This saves the secret from regenerating during an upgrade
+    "helm.sh/hook": "pre-install"
+    "helm.sh/hook-delete-policy": "before-hook-creation"
 type: Opaque
 data:
-  secretsSystem: {{ include "hydra.secrets.system" . | required "Value secrets.system can not be empty!" | b64enc | quote }}
-  secretsCookie: {{ include "hydra.secrets.cookie" . | required "Value secrets.cookie can not be empty!" | b64enc | quote }}
+  # Generate a random secret if the user doesn't give one. User given password has priority
+  secretsSystem: {{ ( include "hydra.secrets.system" . | default ( randAlphaNum 32 )) | required "Value secrets.system can not be empty!" | b64enc | quote }}
```
aeneasr: Is this cryptographically secure?

helm/charts/oathkeeper/values.yaml
```diff
@@ -114,3 +114,7 @@ deployment:
   annotations: {}
 # Configure node affinity
 affinity: {}
+
+# Configures controller setup
+oathkeeper-maester:
```
aeneasr: I think helm convention is camelCase

Demonsthere: [Naming convention](https://helm.sh/docs/chart_best_practices/#chart-names) names with `-` are fine too, as long as the chart resided in a dir with the same name :)

helm/charts/hydra/templates/secrets.yaml
Demonsthere: what do you mean by that?

aeneasr: Are values coming from `randAlphaNum` cryptographically secure? Meaning that they come from e.g. `/dev/urandom` as opposed to a PRNG

Demonsthere: Took me a while but found the [docs](https://github.com/Masterminds/sprig/blob/40a817194fa6bb2cd370b302c30b27d3706f122d/docs/strings.md#randalphanum-randalpha-randnumeric-and-randascii) for this function, seems we are safe :)

aeneasr: Awesome!

aeneasr:
I think this looks pretty solid! Maybe add some docs to oathkeeper and hydra to show the new changes?

Demonsthere:
> This looks really solid!! My last thought is if we should rename the config setting `oathkeeper-maester.enabled` to `oathkeeper.maester.enabled` (so make it a subkey of oathkeeper instead of a top level thing). It would fit in more nicely with the existing config API. Right now, the hyphen stands very lonely as we're not using hyphens anywhere else in the helm charts.

As much as I agree that it stand out like a sore thumb, the key has to reflect the name of the subchart (which is `oathkeeper-maester`). We would need to use some aliases in order to change that

aeneasr:
Wouldn't it be enough to change https://github.com/ory/k8s/pull/37/files#diff-6784805510d68d1677ccaf9f3b5e05c9R4 to:

```
     condition: oathkeeper.maester.enabled
```

?

PR #38, opened 2019-08-06
Make annotations for the hydra deployment configurable 

max-koehler:
## Proposed changes
The current chart wasn't allowing to configure the deployment template's annotations, which becomes a problem in some setups. In order to allow to configure annotations such as Istio's `rewriteAppHTTPProbers` the chart for Hydra needs to provide a way to do so. 
We added a key to the `deployment` section with the name `annotations` and extended the template. The actual implementation is the same as it was already done for the Ingress templates.

## Checklist

### DIFF
--- a/helm/charts/hydra/templates/deployment.yaml
+++ b/helm/charts/hydra/templates/deployment.yaml
@@ -16,10 +16,10 @@ spec:
       labels:
         app.kubernetes.io/name: {{ include "hydra.name" . }}
         app.kubernetes.io/instance: {{ .Release.Name }}
-        {{- with .Values.deployment.annotations }}
-        annotations:
-          {{- toYaml . | nindent 4 }}
-        {{- end }}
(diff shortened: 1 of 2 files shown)


aeneasr:
Awesome, thank you! I think it would make sense to have these keys as subkeys for deployment, similar to the way we [do it with oathkeeper](https://github.com/ory/k8s/blob/master/helm/charts/oathkeeper/values.yaml#L91-L114)

max-koehler:
@aeneasr you think so ;)
https://github.com/ory/k8s/blob/332e22f030f97b5034c31c3f6833dc466f5524ad/helm/charts/hydra/values.yaml#L96-L121

I do think it's implemented exactly as you pointed out. if not then please let me know what exactly you are missing. 

Cheers

max-koehler:
@aeneasr can you check once more?

aeneasr:
Awesome, thank you!

PR #54, opened 2019-09-24
Update YAML k8 file for changes in Kubernetes 1.16

koenmtb1:
Updated the apiVersion to apps/v1 for the Deployment
Set the image to latest and added the selector field.

## Related issue
No GitHub issue but came across this issue while deploying Oathkeeper in Kubernetes.

In the latest version of Kubernetes (1.16) there was a breaking change which moved the Deployment kind out of beta. [https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/](https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/)

## Proposed changes

I changed the apiVersion to apps/v1 instead of extensions/v1beta1. And to prevent another error also added a selector matchLabel.
Another change to make sure everything stays up to date is to change the image version to latest instead of version 0.13.

These changes allow this file to work again.

## Checklist

## Further comments
For now I changed the image to latest so that this file doesn't have to be changed with every release. Feedback on this is welcome.

### DIFF
--- a/yaml/oathkeeper/simple/oathkeeper-api.yaml
+++ b/yaml/oathkeeper/simple/oathkeeper-api.yaml
@@ -30,7 +30,7 @@ spec:
     spec:
       containers:
       - name: ory-oathkeeper
-        image: oryd/oathkeeper:latest
+        image: oryd/oathkeeper:v0.13.8_oryOS.8
         imagePullPolicy: Always
         command: ["oathkeeper", "serve", "api"]
         env:
--- a/yaml/oathkeeper/simple/oathkeeper-api.yaml
+++ b/yaml/oathkeeper/simple/oathkeeper-api.yaml
@@ -12,12 +12,15 @@ spec:
     name: http-ory-oathkeeper
     targetPort: http-api
 ---
-apiVersion: extensions/v1beta1
+apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: ory-oathkeeper
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      app: ory-oathkeeper
   strategy:
     type: RollingUpdate
   template:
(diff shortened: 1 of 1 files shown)


yaml/oathkeeper/simple/oathkeeper-api.yaml
```diff
@@ -27,7 +30,7 @@ spec:
     spec:
       containers:
       - name: ory-oathkeeper
-        image: oryd/oathkeeper:v0.13.8_oryOS.8
+        image: oryd/oathkeeper:latest
```
aeneasr: While this makes sense from an outward perspective, this chart is really only being used by a tutorial we wrote. That tutorial specifically targets this version and is outdated for the actual latest versions. The same applies to the command etc. So this definitely needs to be reverted

```suggestion
        image: oryd/oathkeeper:v0.13.8_oryOS.8
```

koenmtb1: I also noticed that that tutorial is pretty outdated. Isn't it an idea to write a new up-to-date one that reflects the current situation and also make a chart that is always up to date (and can be installed without helm)? I'd be willing to write that (mostly for the fact that I had quite a few troubles while deploying a istio-ambassador-ORY stack).

aeneasr: Absolutely! This has been on my agenda for so long but I simply lack time to do it. It would be awesome if you could help out and I'd be happy to help you along the way.

If you'd like, I can send over the markdown source file for the article?

koenmtb1: Sure! I'm in the discord under the same username, or send through GitHub. And where do I send/merge it once it's finished?
I'll change this version back to v0.13.8 and make a new chart for the new tutorial.

aeneasr:
Thank you!

PR #67, opened 2019-10-18
Hydra deployment serviceaacount

venturasr:
## Related issue
In order to deploy Hydra in our current cluster setup, it is needed to define a PodSecurityPolicy.

This PodSecurityPolicy has to be bound to a ServiceAccount. However currently the hydra/templates does not provide it.

This pull request contribute to the helm chart, specifying a ServiceAccount on the hydra deployment, which is also a good practice.

This pull request will help to tackle possible PodSecurityPolicy related issues like the one below.

```message: 'pods "hydra-966ffcbf4-qzq6w" is forbidden: unable to validate against any pod security policy: [spec.initContainers[0].securityContext.runAsNonRoot: Invalid value: false: must be true]' ```

## Proposed changes

This PR adds an rbac to hydra and defines a ServiceAccount for hydra deployment.

## Checklist

### CODE REVIEW

File: helm/charts/hydra/templates/rbac.yaml [hunk 1]
```diff
@@ -0,0 +1,32 @@
...
+  namespace:  {{ .Release.Namespace }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "hydra.fullname" . }}
+  namespace:  {{ .Release.Namespace }}
+rules:
+  - apiGroups: ["*"]
+    resources: ["pods", "deployments", "services", "configmaps", "serviceaccounts"]
```
aeneasr: Why do we need access to pods, deployments, services, configmaps and serviceaccounts? And why list, get, watch, create, update, patch, delete? Wouldn't it be much better to just have a very minimalistic service account? It seems like this permission list is very exhaustive and not neccessary for hydra to work

helm/charts/hydra/templates/rbac.yaml
```diff
@@ -0,0 +1,32 @@
...
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: {{ include "hydra.fullname" . }}
+  namespace: {{ .Release.Namespace }}
+subjects:
+  - kind: ServiceAccount
+    name: {{ include "hydra.fullname" . }}
+    namespace:  {{ .Release.Namespace }}
+---
```
aeneasr: Add newline

helm/charts/hydra/templates/rbac.yaml
venturasr: Hi @aeneasr 

Yes, but with Role the serviceaccount will have effect at a namespace level which is less than a ClusterRole. However I agree we should reduce the number of verbs.

For this particular use-case where we have PodSecurityPolicy I think this should be the possible rules:
```
- apiGroups: ["*"]
  resources: ["pods", "deployments"]
  verbs: ["get", "list", "watch", "create", "update"]
- apiGroups: ["*"]
  resources: [serviceaccounts]
  verbs: ["get", "list", "watch"]
```

venturasr: Hi @aeneasr 

Sure you followed the conversation in the public slack channel. Here is a particular scenario involving istio  and where we have to define a PodSecurityPolicy. Actually we define PodSecurityPolicy in all our kuberntes applications deployed in our cluster not only hydra.

aeneasr: I believe it would make more sense to have the service account name configurable, that way you can "mount" any service account you'd like. As said before, the "raw" pods of Hydra do not need any of this, so - imo - this shouldn't be here.

aeneasr:
Awesome - thank you for your contribution!

Issue #76, opened 2019-10-29
hydra maester: DNS resolution fails when using fullnameOverride


paulbdavis:
**Describe the bug**

When using a fullname override that contains multiple "parts" it can cause the generated name for the admin service to fail

**To Reproduce**

Deploy the chart with the following name overrides included with the required values:

```yaml
fullnameOverride: foo-ory-hydra
hydra-maester:
  fullnameOverride: foo-ory-hydra-maester
```

**Expected behavior**

I would expect to the template in `_helpers.yaml` to first check for `.Values.adminService` for setting up the hostname and port, for example

```yaml
adminService:
  name: foo-ory-hydra-admin
  port: 9999
```

(the port number is not an issue here, as I am using the defaults, but it should not be hard coded into the maester subchart anyway, in case custom values are used for `.Values.service.admin.port` in the main chart)

Perhaps the maester stuff should be in the same chart rather than a sub chart so that these values can be shared (unless you can grab main chart values from a subchart? not totally sure)?

**Environment**

chart version 0.0.42

aeneasr:
> Perhaps the maester stuff should be in the same chart rather than a sub chart so that these values can be shared (unless you can grab main chart values from a subchart? not totally sure)?

Yes, I believe you can fetch this in subcharts: https://v3.helm.sh/docs/topics/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

aeneasr:
I believe this has been resplved

PR #77, opened 2019-10-29
WIP: Combine hydra and hydra-maester charts

paulbdavis:
## Related issue

#76 

## Proposed changes

Allows for the admin service to be specified rather than auto generated

## Checklist


helm/charts/hydra/charts/hydra-maester/values.yaml
```diff
@@ -14,6 +14,10 @@ image:
   # Image pull policy
   pullPolicy: IfNotPresent
 
+adminService:
```
aeneasr: I believe there is no need to define additional keys for this, see: https://v3.helm.sh/docs/topics/chart_template_guide/subcharts_and_globals/#overriding-values-from-a-parent-chart

aeneasr: Oh, I didn't know that - I thought the parent chart is the "global" value per default. Thank you for clarifying that!

An alternative would be to use a template for this, as templates can be referenced directly from sub-charts: https://helm.sh/docs/chart_template_guide/#sharing-templates-with-subcharts

That way, there would be no extra configuration required for the maester chart, it would just work if enabled - which would be super awesome!

paulbdavis: The issue with sharing templates is that you still have to pass values into the template (the `.` in `{{ include "hydra-masetr.adminService.name" . }}`). The template will still render but it will use the values passed in. Since we don't have access to the parent chart values, we can't pass them into the templates either, so calling `{{ include "hydra.fullname" . }}` from the sub chart would use the `fullnameOverride` from the sub chart values, not the parent chart.

As far as I can tell from my research, the only options are to either have the keys defined in the values or to combine the two applications into a single chart rather than a parent/sub chart.

aeneasr: Damn! Ok, thank you for being so thorough. I think combining charts could be a solution, I think the less we have to configure to get things working, the better for the end user - as long as all functionality is still accessible.

I can imagine a case where a developer changes the port to e.g. 80 and forgets to set up the other services correctly - it will take some time to debug this.

What do you think about merging the charts?

paulbdavis: I think that is probably the best solution, as you'd likely never need to deploy the maester without the main chart. I can work this up and I'll update the PR

paulbdavis: I am going to be deploying oathkeeper as well soon, should I give that chart the same treatment?

aeneasr: Awesome - thank you paul! Yes that would be very useful :)

paulbdavis: It looks like the CRD for the version of the maester running is outdated for the latest version of the maester (which is tagged for oryOS11, but the 1.0.0 for hydra is oryOS12)

Should the versions in the charts be updated to the latest versions as a part of this?

aeneasr: yes!

paulbdavis:
I've pushed up the initial version of the combined hydra chart for review. It's still using my image for the maester image until there is a tagged release that uses the the new CRD and the `--forwarded-proto` flag.

This would supercede #86 as the new CRD is included in this PR. It does use the `v1alhpa1` version until the question of updating that is resolved as well in ory/hydra-maester#30

paulbdavis:
So I am looking at some issues running multiple masters within a cluster. I am working up a patch to address that, but it would mean that these should actually remain separate charts and actually not use a subchart.

aeneasr:
> So I am looking at some issues running multiple masters within a cluster. I am working up a patch to address that, but it would mean that these should actually remain separate charts and actually not use a subchart.

Is the problem that you'd usually only deploy one CRD controller (with one config) instead of multiple ones because all of them would then try to access the annotations and execute some logic?

paulbdavis:
> > So I am looking at some issues running multiple masters within a cluster. I am working up a patch to address that, but it would mean that these should actually remain separate charts and actually not use a subchart.
> 
> Is the problem that you'd usually only deploy one CRD controller (with one config) instead of multiple ones because all of them would then try to access the annotations and execute some logic?

That's correct. In the separated version, the `hydra-maester` chart is still listed as a dependency of the `hydra` chart (as mong as `maester.enabled` is true) so that a single instance deployment can still be done with one chart, but that will require adding the `adminService` stuff to `hydra-maester` as originally proposed here

aeneasr:
> That's correct. In the separated version, the hydra-maester chart is still listed as a dependency of the hydra chart (as mong as maester.enabled is true) so that a single instance deployment can still be done with one chart, but that will require adding the adminService stuff to hydra-maester as originally proposed here

Ok, then there's no other way to solve this. Thank you so much with trying out all of the crazy ideas in here and keeping on churning out those commits! You can imo happily revert the PR (this or a new one) to the original!

Issue #75, opened 2019-10-28
hydra: DSN should be excluded from the configmap


paulbdavis:
**Describe the bug**

The DSN definition is making it's way into the configmap while the value is being consumed via the secretRef env configuration. This is in the hydra chart (though I have not tested the other charts yet)

**To Reproduce**

Install the chart as normal

**Expected behavior**

I would expect the DSN to only be stored in the secret that is set up

**Environment**

chart version 0.0.42

aeneasr:
Thank you for the report, closing because this is a dupe of: #47

Issue #47, opened 2019-08-23
helm/hydra: rework configMap and how secrets are handled


max-koehler:
**Is your feature request related to a problem? Please describe.**
The generation of secrets in the current helm chart are not secure, as the values are stored in plaintext as part of the configMap. 

In order to provide values for `dsn`, `secrets.system` and `secrets.cookie` someone has to set or provide values for the matching config keys such as `hydra.config.dsn`. Later in the `_helpers.tpl` the `hydra.dsn`(as an example here) will be propagated with the value of the provided key `hydra.config.dsn`. 

Once they are provided and part of the config structure they aren't removed at any time and end up in the configMap in plaintext. Eventually the configMap will be mounted as a volume for the application. 

As a result we end up having secrets we inject via environment variables in a secure way, while we are still mounting the same keys/values in plain text through the configMap.

**Describe the solution you'd like**
Remove secrets from the config structure and allow to set the keys directly:.
`hydra.config.secrets.system` -> `hydra.secrets.system`
`hydra.config.secrets.cookies` -> `hydra.secrets.cookie` 
`hydra.config.dsn` -> `hydra.dsn`
Also if those keys are provided are throughout the config object, those keys and it's values should be removed from it before the configMap is stored. 

As this is a breaking change in the current implementation, id like to get your input first but would be willing to bring the changes in as a PR.

aeneasr:
Oh I see - we actually provide this using secrets here https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/secrets.yaml#L15-L17 and generate them from `hydra.config...` ( https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/_helpers.tpl#L70-L78 ). However, I forgot to truncate those values from the actual configmap, which is why they're leaking. If you are willing to supply a fix for that, it would be super appreciated!

evalsocket:
Hey @aeneasr, I like to pick that issue....i think changes are in Line 76 https://github.com/ory/k8s/blob/728fc908cf04a3246b50409b35a8dfb8d7dd8bbd/helm/charts/hydra/templates/_helpers.tpl#L76

aeneasr:
@evalsocket awesome! Unfortunately not, what we have to do is truncate/delete `hydra.config.secrets` as well as `hydra.config.dsn` before rendering it to the configmap. I'm not sure if that possible with helm templates. The config map is being rendered here:

[...]

PR #78, opened 2019-10-29
redact dsn and secrets from config when setting up ConfigMap

paulbdavis:
## Related issue

#47 #75 

## Proposed changes

This replaces the `dsn` property in the config object with "redacted" when setting up the ConfigMap for the other settings

## Checklist

### DIFF
--- a/helm/charts/hydra/templates/configmap.yaml
+++ b/helm/charts/hydra/templates/configmap.yaml
@@ -8,4 +8,5 @@ metadata:
 data:
   "config.yaml": |
   {{- $config := set .Values.hydra.config "dsn" "redacted" -}}
+  {{- $config := set .Values.hydra.config "secrets" "redacted" -}}
   {{- toYaml $config | nindent 4 }}
--- a/helm/charts/hydra/templates/configmap.yaml
+++ b/helm/charts/hydra/templates/configmap.yaml
@@ -7,6 +7,4 @@ metadata:
 {{ include "hydra.labels" . | indent 4 }}
 data:
   "config.yaml": |
-  {{- $config := set .Values.hydra.config "dsn" "redacted" -}}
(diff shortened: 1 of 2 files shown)


helm/charts/hydra/templates/configmap.yaml
```diff
@@ -7,6 +7,6 @@ metadata:
 {{ include "hydra.labels" . | indent 4 }}
 data:
   "config.yaml": |
-  {{- with .Values.hydra.config -}}
-    {{- toYaml . | nindent 4 }}
-  {{- end -}}
+  {{- $config := set .Values.hydra.config "dsn" "redacted" -}}
```
aeneasr: I think we could use [`unset`](https://github.com/Masterminds/sprig/blob/master/docs/dicts.md#unset) instead of set.

paulbdavis:
> Thank you for tackling this! I think this is step in the right direction but I have some feedback:
> 
>     1. I believe that `set|unset` must be a dict for set/unset to work, see: https://github.com/Masterminds/sprig/blob/master/docs/dicts.md#unset
> 
>     2. I believe that `$config` should be created/set in the `_helpers` file, similar to: https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/_helpers.tpl#L30-L32

Not sure what you mean with 1, the config property is an object, which I believe is being treated correctly as a dict for sprig's purposes (the resulting yaml from rendering the chart correctly replaces it).

For 2, I'll get it all put into a template in `_helpers.yaml`

aeneasr:
> Not sure what you mean with 1, the config property is an object, which I believe is being treated correctly as a dict for sprig's purposes (the resulting yaml from rendering the chart correctly replaces it).

Maybe that's just my understanding of things, I haven't actually tried it out myself. I'll do so today though to confirm everything :)

aeneasr:
Thank you, I just checked this out locally and it works like a charm!

PR #87, opened 2019-11-11
Separated hydra-maester chart

paulbdavis:
## Related issue

Related to #76 and ory/hydra-maester#35 

## Proposed changes

This separates the hydra-maester chart to allow it to be deployed separately so that a single hydra-maester instance can manage several hydra instances in a cluster.

This incorporates changes from #77 for the configuration of the (now default) hydra admin instance.

This reqires the changes in ory/hydra-maester#35 to be merged and tagged (temporary images that I have build are in the chart as of the opening of this PR)

The hydra chart still lists the hydra-maester chart as a dependency that is deployed be default. To use the new behavior, `maester.enabled` must be set to false.

## Checklist


helm/charts/hydra/values.yaml
```diff
@@ -147,3 +147,15 @@ affinity: {}
 # Configures controller setup
 maester:
   enabled: true
+
+# Values for the hydra-maester chart
+hydra-maester:
```
aeneasr: Would it be possible to move this under `master` and `config` or does this have to reflect the subchart name?

helm/charts/hydra/requirements.yaml
```diff
@@ -1,4 +1,4 @@
 dependencies:
-  - name: hydra-maester
-    version: 0.0.1
+  - name: ory.sh/hydra-maester
```
aeneasr: Is this intended? The chart name is `hydra-maester` right?

helm/charts/hydra/values.yaml
paulbdavis: It does not have to match the chart name directly, but it does have to be contained in a single object. It could be renamed to `maester` and `maester.enabled` can still be checked but the subdoc has to be together as one

helm/charts/hydra/requirements.yaml
paulbdavis: Yes, since it's no longer directly included in the `charts/` directory, it needs the repo name added so helm can pull it down when installing

aeneasr:
I lost a bit track of the PRs - I think this one still needs to be merged right? If so, could you please resolve the merge conflicts? :)

aeneasr:
This fails currently with:

```
Error: found in requirements.yaml, but missing in charts/ directory: ory.sh/hydra-maester
```

helm/charts/hydra/requirements.yaml
aeneasr: Not sure, this appears to be causing:

```
Error: found in requirements.yaml, but missing in charts/ directory: ory.sh/hydra-maester
```

helm/charts/hydra/requirements.yaml
```diff
@@ -1,4 +1,5 @@
...
```
aeneasr:
```suggestion
  - name: hydra-maester
```

Issue #91, opened 2019-11-22
[helm] maester uses wrong service name 


ms42Q:
### Environment

I am using the hydra chart as a requirement in a project.

```
dependencies:
  - name: hydra
    version: 0.0.46
    repository: "@ory"
```

### Issue

When I install my helmchart with a random release name, maester can't reach the hydra-admin service.

maester logs:
```
ERROR	controller-runtime.controller	Reconciler error	{"controller": "oauth2client", "request": "default/fashionable-kangaroo-user-panel", "error": "Post http://fashionable-kangaroo-admin:4445/clients: dial tcp: lookup fashionable-kangaroo-admin on 10.43.0.10:53: no such host"}
```
### Cause of the issue
There is obviously a mismatch between the service name and the name maester is trying to lookup:
```
-> # kubectl get services
fashionable-kangaroo-hydra-admin                  ClusterIP   10.43.218.112   <none>        4445/TCP   3m20s
fashionable-kangaroo-hydra-public                 ClusterIP   10.43.22.59     <none>        4444/TCP   3m20s
```
Note that in the maester configuration the hydra url  is missing the `-hydra` after `fashionable-kangaroo` as the following output illustrates:
```
 -> # kubectl get deployment fashionable-kangaroo-hydra-maester -o json
.......
            "spec": {
                "containers": [
                    {
                        "args": [
                            "--metrics-addr=127.0.0.1:8080",
                            "--hydra-url=http://fashionable-kangaroo-admin",
                            "--hydra-port=4445"
                        ],
........
```
### Workaround

If the helm release name ends with `-hydra`, maester works. But in that case all the generated names of the hydra resources are not as I expected. I hope the following examples make it clear.

Helm release name **not** ending with `-hydra` (fashionable-kangaroo):
```
-> # kubectl get deployments
NAME                                              READY   UP-TO-DATE   AVAILABLE   AGE
fashionable-kangaroo-hydra                        1/1     1            1           4m43s
fashionable-kangaroo-hydra-maester                1/1     1            1           4m43s
-> # kubectl get services
NAME                                              TYPE        CLUSTER-IP      EXTERNAL-IP   PORT(S)    AGE
fashionable-kangaroo-hydra-admin                  ClusterIP   10.43.218.112   <none>        4445/TCP   3m20s
fashionable-kangaroo-hydra-public                 ClusterIP   10.43.22.59     <none>        4444/TCP   3m20s
```

Helm release name ending with `-hydra` (random-hydra):

```
[...]

PR #92, opened 2019-11-25
Fix hydra admin service name generator

ms42Q:
## Proposed changes

To understand this change, one must know that the hydra chart doesn't include `hydra` in its resource name twice if it is already contained in the release name. This behaviour is defined in: `/helm/charts/hydra/templates/_helpers.tpl:14-22`
```
{{- define "hydra.fullname" -}}
{{- if .Values.fullnameOverride -}}
{{- .Values.fullnameOverride | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- $name := default .Chart.Name .Values.nameOverride -}}
{{- if contains $name .Release.Name -}}
{{- .Release.Name | trunc 63 | trimSuffix "-" -}}
{{- else -}}
{{- printf "%s-%s" .Release.Name $name | trunc 63 | trimSuffix "-" -}}
```
A release name like `hydra-test`, `test-hydra` or simply `hydra` results in the admin service ingress name `<Release.Name>-admin`. A realese name without hydra results in `<Release.Name>-hydra-admin`. So hydra's behaviour observed in #91 is intended and works fine.

Note that the name can be completly overriden using a local value <.Values.fullnameOverride>. If this variable is set, the resource name will be `<.Values.fullnameOverride>-admin`.

My conclusion was that the hydra-maester helmchart needs to reflect this behaviour in order to adjust it's confg. And indeed the _helpers file includes a function `Get hydra admin service name`. The function that is present doesn't reflect this behaviour though.

To fix the issue, this function needed to be changed.

Note that I added an additional Value called `hydraFullnameOverride` with which we can propagate the override of the fullname. in a values.yaml that would look like:
```
hydra:                                                                                         
  fullnameOverride: test                                                                       
  hydra-maester:                                                                               
    hydraFullnameOverride: test         
```

I tested the following release names:

```
hydra,
hydra-test
test-hydra
test-hydra (with fullnameOverride and hydraFullnameOverride set to test)
```
All worked as expected.

## Checklist

### DIFF
--- a/helm/charts/hydra/charts/hydra-maester/templates/_helpers.tpl
+++ b/helm/charts/hydra/charts/hydra-maester/templates/_helpers.tpl
@@ -49,11 +49,11 @@ app.kubernetes.io/managed-by: {{ .Release.Service }}
 Get Hydra admin service name
 */}}
 {{- define "hydra-maester.adminService" -}}
-{{- $fullName := include "hydra-maester.fullname" . -}}
-{{- $nameParts := split "-" $fullName }}
-{{- if eq $nameParts._0 $nameParts._1 -}}
-{{- printf "%s-admin" $nameParts._0 | trimSuffix "-" -}}
+{{- if .Values.hydraFullnameOverride -}}
+{{- printf "%s-admin"  .Values.hydraFullnameOverride -}}
+{{- else if contains "hydra" .Release.Name -}}
+{{- printf "%s-admin" .Release.Name -}}
 {{- else -}}
-{{- printf "%s-%s-admin" $nameParts._0 $nameParts._1 | trimSuffix "-" -}}
+{{- printf "%s-%s-admin" .Release.Name "hydra" -}}
 {{- end -}}
 {{- end -}}


paulbdavis:
This is resolved in #87 as well by making it more explicit when defining a "default" hydra service (that PR also makes changes for the updated CRD that allows specifying a hydra instance per OAuth2Client)

aeneasr:
Right - if you could rebase #87 onto master I'll merge it right away :)

paulbdavis:
@aeneasr rebased and updated the image to the offical repo again now that it's tagged

ms42Q:
Cool. afaic #87 doesn't touch the helpers template so they PRs don't really overlap.
You might want to use my suggested change to set default values for the hydra-admin url.

In any case, let me know how I can help solving this. I really want to use maester as soon as possible.

aeneasr:
I checked, I believe they overlap because you can now set the admin name (if e.g. fullNameOverride is set): https://github.com/ory/k8s/pull/87/files#diff-331de529c7b9c158643f30e37306d9fcR154

Correct me if I'm wrong :)

ms42Q:
Not sure. The current version doesn't work if you use a release name with 2 hypehens.
because of this line:

```
{{- printf "%s-%s-admin" $nameParts._0 $nameParts._1 | trimSuffix "-" -}}
```

aeneasr:
I might be mistaken but if set, no suffix-shuffling is used:

https://github.com/ory/k8s/pull/95/files#diff-2c2ba33f95125b4cba27abe5e984fd06R52-R54

ms42Q:
Exactly. You are right about the override.

I was refering to the case where you don't specify the override. in https://github.com/ory/k8s/pull/95/files#diff-2c2ba33f95125b4cba27abe5e984fd06R52-R54 that would be the {{else}} case.

with test-test as a release name you end up with
test-admin when it should be `test-test-hydra-admin`

ms42Q:
Originally i created this PR not because of the override. I wasn't even aware that it existed. I just added the override case for completeness.

aeneasr:
Right, it makes sense the way you've set it up here. Thank you!

PR #95, opened 2019-11-26
Separate hydra-maester into subchart (#87)

aeneasr:
Related to #76 and ory/hydra-maester#35

This separates the hydra-maester chart to allow it to be deployed separately so that a single hydra-maester instance can manage several hydra instances in a cluster.

This incorporates changes from #77 for the configuration of the (now default) hydra admin instance.

This reqires the changes in ory/hydra-maester#35 to be merged and tagged (temporary images that I have build are in the chart as of the opening of this PR)

The hydra chart still lists the hydra-maester chart as a dependency that is deployed be default. To use the new behavior, maester.enabled must be set to false.

## Checklist

- [ ] I have read the [security policy](../security/policy)
- [ ] I confirm that this pull request does not address a security vulnerability. If this pull request addresses a security


helm/charts/hydra/requirements.yaml
```diff
@@ -1,4 +1,5 @@
 dependencies:
-  - name: hydra-maester
-    version: 0.0.1
+  - name: ory.sh/hydra-maester
```
aeneasr:
```suggestion
  - name: hydra-maester
```

helm/charts/hydra/requirements.yaml
```diff
@@ -1,4 +1,5 @@
 dependencies:
   - name: hydra-maester
```
aeneasr:
```suggestion
  - name: ory/hydra-maester
```

aeneasr:
@paulbdavis it's apparently quite tricky to get subcharts to work when they are in the parent directory, so I moved it back to the original place. I've added the CI chain so that hydra-maester is now being released on its own, from the original directory. I think this will resolve all issues with testing, etc.

Issue #96, opened 2019-11-27
Installation fails with helm3 on minikube


skuro:
**Describe the bug**

The `ory/hydra` chart fails to install under `helm3`

**To Reproduce**

1. install [helm3](https://github.com/helm/helm/releases/tag/v3.0.0)
2. `helm repo add ory https://k8s.ory.sh/helm/charts`
3. `helm repo update`
4. try to run `helm install` as per the [official guide](https://k8s.ory.sh/helm/hydra.html)
5. the installation fails with `Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(RoleBinding.roleRef): unknown field "namespace" in io.k8s.api.rbac.v1beta1.RoleRef`

**Expected behavior**

The helm chart is installed successfully

**Environment**

* Version: v1.0.0-rc.14_oryOS.12
* Environment: helm v3.0.0, minikube v1.5.2 on Ubuntu

aeneasr:
Helm3 is very new, please use Helm2 for the time being when working with these charts.

skuro:
thanks for the tip @aeneasr, although I actually already did that :-) Still thought it was good to track this one down with its own issue

matdehaast:
This can be worked around by disabling the use of the maester. Use the following flag when installing the chart

```
 --set 'maester.enabled=false' 
```

skuro:
good to know, thanks @matdehaast

PR #97, opened 2019-12-03
Remove namespace as unsupported feature

kosfist:
Remove namespace as unsupported feature for roleRef.
This change must fix error installing hydra via helm3.

## Related issue
#96


aeneasr:
Thank you and sorry for the long review delay! In fact I thought I already merged a PR for this, but apparently not :D

PR #125, opened 2020-03-30
chore: Allow usage of existing secret in hydra chart

clement-buchart:
Signed-off-by: Clément BUCHART <clement@buchart.dev>

## Proposed changes

Simply allow using an existing secret to configure hydra.
Our use-case is as follow : 
We create a Google CloudSQL and a gcloud-sqlproxy via terraform while setting up our infrastructure.
We deploy Hydra as part of our application stack.

We would like to be able to have terraform create the secrets containing the DataSourceName as part of the infrastructure setup, and have hydra pick the secret up during the application deploy step.

## Checklist

### DIFF
--- a/docs/helm/hydra.md
+++ b/docs/helm/hydra.md
@@ -13,7 +13,7 @@ To install ORY Hydra, the following values must be set
 * `hydra.config.urls.consent`
 * `hydra.config.secrets.system`
 
-> **NOTE:** If no `hydra.config.secrets.system` secrets are not supplied, a secret is generated automatically. The generated secret is cryptographically secure, and 32 signs long.
+> **NOTE:** If no `hydra.config.secrets.system` secrets is supplied and `hydra.existingSecret` is empty, a secret is generated automatically. The generated secret is cryptographically secure, and 32 signs long.
 
 If you wish to install ORY Hydra with an in-memory database, a cryptographically strong secret, a Login and Consent
 provider located at `https://my-idp/` run:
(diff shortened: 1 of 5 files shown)


helm/charts/hydra/values.yaml
```diff
@@ -79,6 +79,8 @@ ingress:
 #        - api.hydra.local
 #      - secretName: hydra-api-example-tls
 
+# Use a pre-existing secret
+# existingSecret: my-preexisting-secret
```
aeneasr: I think it makes sense to move this to `hydra` section (like `autoMigrate`). What do you think? Also, to make it maybe a bit clearer what this does, what do you think about the name `secretKeyRef`?

clement-buchart: Yes, I wasn't sure about the proper location. Will move it as per your suggestion.

Not sure about `secretKeyRef` though.
I would expect a `secretKeyRef` to refer to a secret _key_ (such as _secretsCookie_ or _secretsSystem_, not a secret _name_)

I went with existingSecret because it is used in a [fair amount of helm charts in the stable repository](https://github.com/helm/charts/search?q=secret+existing&unscoped_q=secret+existing) which makes it a convention of sorts.

aeneasr: Oh, you're absolutely right! This looks good to me :)

aeneasr:
Thank you, good job :)

docs/helm/hydra.md
```diff
@@ -13,7 +13,7 @@ To install ORY Hydra, the following values must be set
 * `hydra.config.urls.consent`
 * `hydra.config.secrets.system`
 
-> **NOTE:** If no `hydra.config.secrets.system` secrets are not supplied, a secret is generated automatically. The generated secret is cryptographically secure, and 32 signs long.
+> **NOTE:** If no `hydra.config.secrets.system` secrets is supplied and `existingSecret` is empty, a secret is generated automatically. The generated secret is cryptographically secure, and 32 signs long.
```
aeneasr:
```suggestion
> **NOTE:** If no `hydra.config.secrets.system` secrets is supplied and `hydra.existingSecret` is empty, a secret is generated automatically. The generated secret is cryptographically secure, and 32 signs long.
```

clement-buchart: Oups, fixed

aeneasr:
Thank you, great work!

Issue #134, opened 2020-04-24
adminService in hydra's values yaml causes warning


bjayanth:
**Describe the bug**
While installing hydra as a dependency with Helm3, there is a warning:
```
coalesce.go:196: warning: cannot overwrite table with non table for adminService (map[name:<nil> port:<nil>])
```

**To Reproduce**
Have a parent chart that depends on hydra-0.3.3 Helm chart, override some values, like the fullnameOverride.

Chart.yaml:
```
dependencies:
  - name: hydra
    version: 0.3.3
    repository: https://k8s.ory.sh/helm/charts
```

values.yaml:
```
hydra:
  fullnameOverride: &HYDRA_FULLNAME "release-hydra"
  maester:
    hydraFullnameOverride: *HYDRA_FULLNAME
    fullnameOverride: hydra-maester
```
**Expected behavior**

There is no warning while installing hydra as a dependency.

**Environment**

- Helm Version: `version.BuildInfo{Version:"v3.1.2", GitCommit:"d878d4d45863e42fd5cff6743294a11d28a9abce", GitTreeState:"clean", GoVersion:"go1.13.8"}`

**Possible Fix**
Having `adminService` as a map removes the warning.
```
# Configures controller setup
maester:
  enabled: true
  # Values for the hydra admin service arguments to hydra-maester
  adminService: {}
```

aeneasr:
Would you be open to give it a shot? :)

PR #135, opened 2020-04-24
make maester.adminService in hydra values.yaml a map to remove warning

bjayanth:
make maester.adminService in hydra values.yaml a map to remove coalesce warning

## Related issue

https://github.com/ory/k8s/issues/134

## Proposed changes

maester.adminService is map with two values (name and port). Not having `{}` causes a warning.

```
# Configures controller setup
maester:
  enabled: true
  # Values for the hydra admin service arguments to hydra-maester
  adminService: {}  # -----> instead of adminService:
    # The service name value may need to be set if you use
    # `fullnameOverride` for the parent chart
    # name:

    # You only need to set this port if you change the value for
    # `service.admin.port` in the parent chart
    # port:

```

## Checklist

      the changes.

### DIFF
--- a/helm/charts/hydra/values.yaml
+++ b/helm/charts/hydra/values.yaml
@@ -149,7 +149,7 @@ affinity: {}
 maester:
   enabled: true
   # Values for the hydra admin service arguments to hydra-maester
-  adminService:
+  adminService: {}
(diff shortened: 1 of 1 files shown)


aeneasr:
Thank you!

PR #144, opened 2020-05-27
upkeep: use GA RBAC resource api version

pjediny:
As of K8s 1.8, RBAC mode is stable and backed by the
rbac.authorization.k8s.io/v1 API


aeneasr:
thank you!

Issue #153, opened 2020-06-21
Kratos chart doesn't insist on all the required configuration


simeonmiteff:
**Describe the bug**

By careful eyeballing of the kratos config schema, I've figured out that 0.3.0 requires:
- `kratos.config.dsn`
- `kratos.config.identity.traits.default_schema_url`
- `kratos.config.courier.smtp.connection_uri`
- `kratos.config.urls.self.public`
- `kratos.config.urls.self.admin`

It doesn't seem like the chart enforces presence of the latter 4 keys in the configuration. The [chart documentation](https://k8s.ory.sh/helm/kratos.html) says the two required values are:
- `kratos.config.dsn`
- `kratos.config.secrets.session`

I have opened #152 to clarify whether `kratos.config.secrets.session` is really needed.

**To Reproduce**

1. Follow the [chart documentation](https://k8s.ory.sh/helm/kratos.html)
2. `helm install`
3. Observe kratos fail to start without the above required configuration

**Expected behavior**

1. Documentation directs the chart user to set the required configuration
2. The chart enforces required configuration

**Environment**

- Version: chart version 0.1.0

aeneasr:
I believe this has been improved since the issue has been opened originally!

Issue #151, opened 2020-06-21
Chart for kratos using old version


simeonmiteff:
**Describe the bug**

The helm chart for kratos is referencing version 0.1.1, which is old.

**To Reproduce**

1. `helm install` kratos
2. Observe old version is deployed

**Expected behavior**

1. `helm install` kratos
2. Have a recent (or stable, once kratos is stable) version deployed. At the moment probably 0.3.0

**Environment**

- Version: helm chart version 0.1.0

PR #154, opened 2020-06-21
Fix kratos version

simeonmiteff:
## Related issue

Address #151

## Proposed changes

- Updates the kratos version from 0.1.1 to 0.3.0, and bumps the chart version from 0.1.0 to 0.2.0

## Checklist

      the changes.

## Further comments

Expect to follow with more PRs if required to address #153.

### DIFF
--- a/helm/charts/kratos/templates/secrets.yaml
+++ b/helm/charts/kratos/templates/secrets.yaml
@@ -9,4 +9,4 @@ type: Opaque
 data:
   dsn: {{ include "kratos.dsn" . | b64enc | quote }}
   # Generate a random secret if the user doesn't give one. User given password has priority
-  secretsSession: {{ include "kratos.secrets.session" . | required "Value kratos.config.secrets.session can not be empty!" | b64enc | quote }}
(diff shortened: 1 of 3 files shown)


helm/charts/kratos/templates/secrets.yaml
```diff
@@ -9,4 +9,4 @@ type: Opaque
 data:
   dsn: {{ include "kratos.dsn" . | b64enc | quote }}
   # Generate a random secret if the user doesn't give one. User given password has priority
-  secretsSession: {{ include "kratos.secrets.session" . | required "Value kratos.config.secrets.session can not be empty!" | b64enc | quote }}
+  secretsSession: {{ include "kratos.secrets.session" . | b64enc | quote }}
```
aeneasr, line 12: Even though the secret is not required in Kratos, it's a bad idea for this helm chart to not require it. Without a fixed secret, you will not be able to scale Kratos nor will you be able to restart it, because your users will all be logged out!

Therefore, I suggest we revert this change:

```suggestion
secretsSession: {{ include "kratos.secrets.session" . | required "Value kratos.config.secrets.session can not be empty!" | b64enc | quote }}
```

PR #156, opened 2020-07-09
fix: update kratos' helm chart to use secrets.cookie

### LINKED ISSUE #151: Chart for kratos using old version

zepatrik:
## Related issue

The current chart is not working with kratos@v0.4.3
closes #154 
closes #151 

## Proposed changes

Replaced the usage of `secrets.session` with `secrets.cookie` as the key was renamed.

## Further comments

Maybe we should also respect `secrets.default` as they can be used as a fallback for `sessions.cookie`.

### DIFF
--- a/helm/charts/kratos/templates/_helpers.tpl
+++ b/helm/charts/kratos/templates/_helpers.tpl
@@ -41,9 +41,9 @@ Generate the dsn value
 {{/*
 Generate the secrets.session value
 */}}
-{{- define "kratos.secrets.session" -}}
-{{- if .Values.kratos.config.secrets.session -}}
-{{- .Values.kratos.config.secrets.session }}
+{{- define "kratos.secrets.cookie" -}}
+{{- if .Values.kratos.config.secrets.cookie -}}
+{{- .Values.kratos.config.secrets.cookie }}
 {{- end -}}
 {{- end -}}
 
--- a/helm/charts/kratos/templates/_helpers.tpl
+++ b/helm/charts/kratos/templates/_helpers.tpl
@@ -39,7 +39,16 @@ Generate the dsn value
 {{- end -}}
 
 {{/*
-Generate the secrets.session value
+Generate the secrets.default value
+*/}}
+{{- define "kratos.secrets.default" -}}
+{{- if .Values.kratos.config.secrets.default -}}
+{{- .Values.kratos.config.secrets.default }}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Generate the secrets.cookie value
 */}}
 {{- define "kratos.secrets.cookie" -}}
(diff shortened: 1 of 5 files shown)


zepatrik:
@rauanmayemir

rauanmayemir:
@zepatrik Thanks, I think I managed to set it all up!
At least now I'm being correctly redirected and seeing the flows.

helm/charts/kratos/templates/_helpers.tpl
```diff
@@ -41,9 +41,9 @@ Generate the dsn value
 {{/*
 Generate the secrets.session value
 */}}
-{{- define "kratos.secrets.session" -}}
-{{- if .Values.kratos.config.secrets.session -}}
-{{- .Values.kratos.config.secrets.session }}
+{{- define "kratos.secrets.cookie" -}}
```
rauanmayemir, line 44: You might also want to change default `secrets.session` in `values.yaml`.

rauanmayemir: Or just remove it altogether. That way, user is forced to explicitly set the config value instead of relying on unsafe defaults.

aeneasr: `secrets.default` is also missing :)

helm/charts/kratos/templates/secrets.yaml
```diff
@@ -8,5 +8,6 @@ metadata:
 type: Opaque
 data:
   dsn: {{ include "kratos.dsn" . | b64enc | quote }}
-  # Generate a random secret if the user doesn't give one. User given password has priority
-  secretsSession: {{ include "kratos.secrets.cookie" . | required "Value kratos.config.secrets.cookie can not be empty!" | b64enc | quote }}
+  # Generate a random secret if the user doesn't give one. User given secret has priority
+  secretsDefault: {{ ( include "kratos.secrets.default" . | default ( randAlphaNum 32 )) | required "Value kratos.config.secrets.default can not be empty!" | b64enc | quote }}
```
aeneasr, line 12: Did you check if `randAlphaNumeric` is a cryptographically secure PRNG?

zepatrik: Honestly I just copied that from the hydra chart.

aeneasr:
Thank you!

PR #190, opened 2020-11-17
feat(kratos): Add support for specifying env vars from a secret

bmonkman:
## Related issue

Related to a discussion in Slack and loosely to this change: https://github.com/ory/x/commit/5b71f1e15fa3264e931d74e58965af73a51c8fc5

## Proposed changes

When setting up Kratos via Helm there are currently secrets in the config that should be passed in via another means.
The case I'm dealing with specifically is `kratos.config.selfservice.methods.oidc.config.providers`, which contains `client_id` and `client_secret` for each provider.
In this case my ideal scenario seems to be to set up secret containing the provider config, and then have that create env vars which Kratos can use to override config (in this case: `SELFSERVICE_METHODS_OIDC`.)
This requires that the secret is created separately, but I think that is beneficial. If it were up to Helm to create the secret then anyone who needed to run the chart would also have to have the contents of the secret. And in my case I'm executing Helm with Terraform, which would require the secrets to go into the Terraform state, and be accessible by anyone who needs to run that TF.

## Checklist

      the changes.
- [x] I have added necessary documentation within the code base (if appropriate).

### DIFF
--- a/docs/helm/kratos.md
+++ b/docs/helm/kratos.md
@@ -57,9 +57,10 @@ Additionally, the following extra settings are available:
 
 - `kratos.autoMigrate` (bool): If enabled, an `initContainer` running `kratos migrate sql` will be created.
 - `kratos.development` (bool): If enabled, kratos will run with `--dev` argument.
-- `secret.enabled` (bool): If `true` (default), a Kubernetes Secret is created (contains `dsn`, `secretsCookie` and `secretsDefault`). Also generates `secertsCookie` and `secretsDefault` unless already set.
-- `secret.nameOverride` (string): Let's you override the name of the secret to be used
+- `secret.enabled` (bool): If `true` (default), a Kubernetes Secret is created (contains `dsn`, `secretsCookie` and `secretsDefault`). Also generates `secretsCookie` and `secretsDefault` unless already set.
+- `secret.nameOverride` (string): Lets you override the name of the secret to be used
 - `ingress.admin.enabled` (bool): If enabled, an ingress is created on admin endpoint
 - `ingress.public.enabled` (bool): If enabled, an ingress is created on public endpoint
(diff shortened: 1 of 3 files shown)


bmonkman:
@aeneasr or @zepatrik this is the final piece of that change I made to config parsing. Mind taking a look?

docs/helm/kratos.md
```diff
@@ -57,9 +57,10 @@ Additionally, the following extra settings are available:
...
-- `secret.enabled` (bool): If `true` (default), a Kubernetes Secret is created (contains `dsn`, `secretsCookie` and `secretsDefault`). Also generates `secertsCookie` and `secretsDefault` unless already set.
-- `secret.nameOverride` (string): Let's you override the name of the secret to be used
+- `secret.enabled` (bool): If `true` (default), a Kubernetes Secret is created (contains `dsn`, `secretsCookie` and `secretsDefault`). Also generates `secretsCookie` and `secretsDefault` unless already set.
+- `secret.nameOverride` (string): Lets you override the name of the secret to be used
 - `ingress.admin.enabled` (bool): If enabled, an ingress is created on admin endpoint
 - `ingress.public.enabled` (bool): If enabled, an ingress is created on public endpoint
+- `deployment.environmentSecretsName` (string): Allows you to set arbitrary environment variables from [a secret containing a list of keys and values](https://kubernetes.io/docs/tasks/inject-data-application/distribute-credentials-secure/#configure-all-key-value-pairs-in-a-secret-as-container-environment-variables). (This secret is not created by the Helm chart)
```
aeneasr, line 64: Given that the others are called `secrets.enabled` and `secrets.nameOverride` wouldn't it make sense to keep this also unter that key - for example `secrets.environmentSecretsName`?

aeneasr (commented):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

Also sorry for the late review!

docs/helm/kratos.md
bmonkman: The stuff under `secrets` is specifically about the secret that the helm chart creates and manages. I felt that nesting it under there would be confusing.

aeneasr: Right, that makes sense! Let's keep it as is then!

aeneasr (approved):

aeneasr:
Awesome 🎉  Thank you for your contribution! :)

Issue #138, opened 2020-04-30
Hydra maester cannot be deployed to multiple namespaces


jbapu:
**Describe the bug**

Trying to deploy Hydra to multiple k8s namespaces results in 
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: existing resource conflict: namespace: , name: maester-role, existing_kind: rbac.authorization.k8s.io/v1beta1, Kind=ClusterRole, new_kind: rbac.authorization.k8s.io/v1beta1, Kind=ClusterRole
```

**To Reproduce**

```
helm install hydra ory/hydra --version 0.3.3 -n namespace1
helm install hydra ory/hydra --version 0.3.3 -n namespace2
```
Results in:
```
Error: rendered manifests contain a resource that already exists. Unable to continue with install: ClusterRole "maester-role" in namespace "" exists and cannot be imported into the current release: invalid ownership metadata; annotation validation error: key "meta.helm.sh/release-namespace" must equal "namespace2": current value is "namespace1"
```
**Expected behavior**

I expect to be able to successfully deploy Hydra into different namespaces without any errors.

**Environment**

- Version:
```
$ helm version
version.BuildInfo{Version:"v3.2.0", GitCommit:"e11b7ce3b12db2941e90399e874513fbd24bcb71", GitTreeState:"clean", GoVersion:"go1.13.10"}
```
```
$ kubectl version
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.5", GitCommit:"20c265fef0741dd71a66480e35bd69f18351daea", GitTreeState:"clean", BuildDate:"2019-10-15T19:16:51Z", GoVersion:"go1.12.10", Compiler:"gc", Platform:"windows/amd64"}
Server Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.2", GitCommit:"59603c6e503c87169aea6106f57b9f242f64df89", GitTreeState:"clean", BuildDate:"2020-01-18T23:22:30Z", GoVersion:"go1.13.5", Compiler:"gc", Platform:"linux/amd64"}
```

Is this intended behavior? Can hydra only be deployed once (maybe use `enabledNamespaces` for multiple namespaces)?

aeneasr:
Yeah that's definitely a bug. You don't need maester, right?

bjayanth:
Good point. I guess I don't..

christian-roggia:
[...]

PR #196, opened 2020-12-07
Fix maester chart naming clashes between hydra & oathkeeper

Atharex:
## Related issue
#138 

@aeneasr this should be pretty easy to solve.

### DIFF
--- a/helm/charts/oathkeeper/Chart.yaml
+++ b/helm/charts/oathkeeper/Chart.yaml
@@ -25,5 +25,5 @@ dependencies:
   - name: oathkeeper-maester
     version: 0.4.11
     condition: maester.enabled
-    alias: maester
+    alias: oathkeeper-maester
     repository: file://../oathkeeper-maester
--- a/helm/charts/hydra/Chart.yaml
+++ b/helm/charts/hydra/Chart.yaml
@@ -24,5 +24,5 @@ dependencies:
   - name: hydra-maester
     version: 0.4.11
     condition: maester.enabled
-    alias: maester
+    alias: hydra-maester
     repository: file://../hydra-maester


aeneasr:
Thank you, this looks great! It looks like the CLA bot is not properly detecting your signature. To fix this, try the following:

```
$ git commit  --amend --author="Author Name <email@address.com>"
```

Ensure that `Author Name` is replaced with your GitHub username (e.g. `aeneasr`) and that the email address is replaced with the email address you have set up in GitHub (e.g. `3372410+aeneasr@users.noreply.github.com`):

```
$ git commit  --amend --author="aeneasr <3372410+aeneasr@users.noreply.github.com>"
```

Once that is done, you can force-push your changes (make sure to push to the correct remote and branch!):

```
$ git push --force <remote> HEAD:<branch>
```

helm/charts/oathkeeper/Chart.yaml
```diff
@@ -25,5 +25,5 @@ dependencies:
   - name: oathkeeper-maester
     version: 0.4.11
     condition: maester.enabled
-    alias: maester
```
aeneasr, line 28: I think this could break something in the internal mapping, not sure. Have you tested this change in Kubernetes?

Atharex: I've installed the hydra & oathkeeper charts both with maester with this change and they run with no problems.
Also the logs do not show any obvious errors, so this seems to work :) 
Though I cannot be sure it is not breaking anything internally, as I did not do extensive testing beyond the mentioned steps

aeneasr: Ok, should be good then! :)

Issue #194, opened 2020-12-01
Horizontal Pod Autoscaler for Hydra Helm Chart


pommelinho:
**Is your feature request related to a problem? Please describe.**

We would like to add an HPA for our Hydra Deployment. However there is currently no HorizontalPodAutoscaler Resource in the Hydra Helm Chart.

**Describe the solution you'd like**

Add an HPA Resource to templates like e.g.

```
{{- if .Values.autoscaling.enabled }}
apiVersion: autoscaling/v2beta2
kind: HorizontalPodAutoscaler
metadata:
  namespace: crm
  name: {{ include "hydra-chart.fullname" . }}
  labels:
    {{- include "hydra-chart.labels" . | nindent 4 }}
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: {{ include "hydra-chart.fullname" . }}
  minReplicas: {{ .Values.autoscaling.minReplicas }}
  maxReplicas: {{ .Values.autoscaling.maxReplicas }}
  metrics:
  {{- if .Values.autoscaling.cpuAverageValue }}
    - type: Resource
      resource:
        name: cpu
        target: 
          type: AverageValue 
          averageValue: {{ .Values.autoscaling.cpuAverageValue }}
  {{- end }}
  {{- if .Values.autoscaling.memoryAverageValue }}
    - type: Resource
      resource:
        name: memory
        target:
          type: AverageValue
          averageValue: {{ .Values.autoscaling.memoryAverageValue }}
  {{- end }}
{{- end }}
```

**Describe alternatives you've considered**

Fork from ory/k8s and add HPA.yaml in own repo.
 
**Additional context**

aeneasr:
Hey! Sorry for the late reply! We can definitely add a pod autoscaler and contributions are highly appreciated!

PR #197, opened 2020-12-09
feat: add HPA for hydra +  mockTlsTermination for example-idp

pommelinho:
## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/hydra/values.yaml
+++ b/helm/charts/hydra/values.yaml
@@ -148,7 +148,7 @@ deployment:
 
   extraVolumes: []
   extraVolumeMounts: []
-  # If yout want to mount external volume
+  # If you want to mount external volume
   # For example, mount a secret containing Certificate root CA to verify database
   # TLS connection.
   # extraVolumes:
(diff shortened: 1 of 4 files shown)


pommelinho:
Added HPA for hydra chart.This is partially related to #194 .
Additionally i added support env MOCK_TLS_TERMINATION in example-idp chart which might be useful so that exampl-idp can communicae with hydra admin api.

helm/charts/hydra/values.yaml
```diff
@@ -163,6 +163,14 @@ deployment:
 # Configure node affinity
 affinity: {}
 
+# Configure HPA
+autoscaling:
+  enabled: true
```
aeneasr, line 168: Can we move this to the "deployment" top level key? Also let's disable this per default!

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good but I have some ideas how to improve it further :)

helm/charts/hydra/values.yaml
pommelinho: sure

aeneasr (approved):
Awesome 🎉 

Thank you for your contribution!

PR #226, opened 2021-02-22
chore(kratos): fix kratos chart values.yaml to align with manifest files

zhming0:
## Proposed changes

This PR fixes Kratos chart's `value.yaml` so it aligns with the actual manifest files. For example, in the `deployment.yaml`, it uses `.Values.resources` but in `value.yaml` it says the `resources` goes under the `deployment` key.

This PR also contains some minor comments change on the `value.yaml` aimed at reducing confusion. 

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

I do wonder if the `autoMigrate` should be true by default. I think this might fix most users interest.

### DIFF
--- a/helm/charts/kratos/values.yaml
+++ b/helm/charts/kratos/values.yaml
@@ -72,6 +72,8 @@ ingress:
 
 kratos:
   development: false
+  # autoMigrate is relying on a simple initContainer mechanism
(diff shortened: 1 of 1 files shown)


aeneasr (approved):
Awesome 🎉 

Thank you for your contribution!

PR #232, opened 2021-03-18
fix(hydra): fix values for maester in helm chart

m1pl:
The values for hydra-maester aren't passed to the child chart. This fixes it.


aeneasr:
Awesome, thank you! 🎉 Your contribution makes Ory better :)

Issue #245, opened 2021-04-14
Update keto chart to reflect new version of the application


Demonsthere:
**Is your feature request related to a problem? Please describe.**

The new release of keto seems not to be backwards compatible with the existing helm chart. 

**Describe the solution you'd like**

The chart is updated to reflect changes in the new keto version

PR #246, opened 2021-04-15
feat: rework keto chart

Demonsthere:
## Related issue
- Implements https://github.com/ory/k8s/issues/245

## Proposed changes
- Update the existing keto chart to be compatible with the new version (0.6.0)

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).


helm/charts/keto/templates/NOTES.txt
```diff
@@ -1,21 +1,20 @@
 1. Get the application URL by running these commands:
-{{- if .Values.ingress.enabled }}
-{{- range $host := .Values.ingress.hosts }}
+{{- if .Values.ingress.read.enabled }}
+{{- range $host := .Values.ingress.read.hosts }}
```
zepatrik, line 3: Should we add write as well? There are now two endpoints instead of just one :thinking:

helm/charts/keto/values.yaml
```diff
@@ -8,7 +8,7 @@ image:
   repository: oryd/keto
   pullPolicy: IfNotPresent
   # Overrides the image tag whose default is the chart appVersion.
-  tag: ""
+  tag: "v0.6.0-sqlite"
```
zepatrik, line 11: I mean without alpha also works but maybe make it most verbose
```suggestion
  tag: "v0.6.0-alpha.1-sqlite"
```

helm/charts/keto/values.yaml
```diff
@@ -38,31 +38,60 @@ securityContext:
...
+    # kubernetes.io/ingress.class: nginx
+    # kubernetes.io/tls-acme: "true"
+    hosts:
+      - host: chart-example.local
+        paths: []
+    tls: []
+    #  - secretName: chart-example-tls
+    #    hosts:
+    #      - chart-example.local
+
```
zepatrik, line 66: Same host and port for both endpoints? I guess this is just an example but maybe still change that to not confuse people.

zepatrik: Or is it possible to have it on the same?

zepatrik (commented):
Not that I am super fluent with helm charts...

helm/charts/keto/values.yaml
Demonsthere: It is possible by setting specific paths, but you're right I will make the example more verbose

aeneasr (approved):

Issue #247, opened 2021-04-16
Run migration containers as jobs not initContainer


aeneasr:
**Is your feature request related to a problem? Please describe.**

Because initContainer scale with the number of containers - meaning we get contestion for migrations. Instead, we should run a "run to completion once" container with exactly 1 pod (no concurrency)

Demonsthere:
Possible solution:
- Single job with a CM mounted. In the CM we have some metadata (migration_occured, migration_date), which is set by the pod only if a migration was successful. 
- Deployments mount the CM and have an initJob that waits until the value in the CM is as expected (red light/green light)

aeneasr:
Single job with CM sounds good!

Demonsthere:
First draft, for now only hydra: https://github.com/ory/k8s/pull/248 will add the rest after some testing

PR #248, opened 2021-04-16
feat: Move migration to separate job, add runtime guards

Demonsthere:
## Related issue
- Fixes https://github.com/ory/k8s/issues/247

## Proposed changes
- Move automigration to a separate job, which runs as a pre-install,upgrade hook. This allows us to perform migration before any replicas of hydra are created. 
- Updated charts:
  - [x] hydra
  - [x] keto
  - [x] kratos

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).


helm/charts/hydra/templates/job-mirgation.yaml
```diff
@@ -0,0 +1,70 @@
...
+        - /bin/sh
+        - -c
+        - |
+          set -ex
+          apk add curl 
+          curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+          chmod +x kubectl
+          mv kubectl /usr/local/bin/kubectl
+          [[ -f /etc/migration/migration.json ]] && echo "Migration was done, nothing to do here" && exit 0
+          hydra migrate sql -e --yes --config /etc/config/config.yaml
```
aeneasr, line 35: The migration job is itself capable of identifying if migrations need to  be applied or not - so no need for workarounds here :)

Demonsthere: Good to know, will simplify the procedure :) What are your thought on adding `kubectl` to our images? Something like `hydra-debug` which is based on alpine and has some additional packages?

piotrmsc: 👍  for me to have an image with kubectl, we will need it in the future as well. I had to use bitnami already while upgrading kube-prom-stack crds :) And also it should be maintained by us meaning security scans done constantly etc

helm/charts/hydra/templates/_helpers.tpl
```diff
@@ -140,3 +140,11 @@ Check overrides consistency
 {{- $local := dict "first" true -}}
 {{- range $k, $v := . -}}{{- if not $local.first -}},{{- end -}}{{- $v -}}{{- $_ := set $local "first" false -}}{{- end -}}
 {{- end -}}
+
+{{/*
+Get alpine image
+*/}}
+{{- define "hydra.getImageAlpine" -}}
```
piotrmsc, line 147: neat : `getAlpineImage` sounds bit better?:D

helm/charts/hydra/templates/job-mirgation.yaml
```diff
@@ -0,0 +1,70 @@
...
+          echo {\"timestamp\": \"$(date "+%Y%m%d-%H%M%S")\",\"completed\": true} > /tmp/migration.json
+          kubectl create configmap {{ include "hydra.fullname" . }}-migration --from-file=/tmp/migration.json
```
piotrmsc, line 37: This `kubectl create` should be replaced with ``--dry-run | kubectl apply``` IMO as job will be executed per install/upgrade and will fail with with already exists error from kubectl.

And since this will be part of helm release (Job) won't it cause troubles as images in jobs are immutable? Ideally it should be a hook but probably the config volume CM is not on a hook as well?

Demonsthere: My idea was to delete the cm if it already exists (so, upgrade scenario), which would block new instances from running too early. 
The job will be on pre-install,pre-uprade hooks, just didn't add them yet to better debug locally ;)

helm/charts/hydra/templates/job-mirgation.yaml
aeneasr: Yeah I guess we could add kubectl to the alpine images? I wouldn't have it in the scratch ones though.

helm/charts/hydra/files/automigrate.sh
```diff
@@ -0,0 +1,40 @@
+set -e
```
aeneasr, line 1: This can be removed right?

helm/charts/hydra/values.yaml
```diff
@@ -139,7 +139,7 @@ deployment:
       - ALL
     readOnlyRootFilesystem: true
     runAsNonRoot: true
-    runAsUser: 1000
+    runAsUser: 100
```
aeneasr, line 142: What's the difference?

aeneasr (approved):
This looks good to me!

helm/charts/hydra/files/automigrate.sh
piotrmsc: actually IMO that's good as we want to exit if any command return non 0. 
But I would add here also `-o pipefail` as we have some pipes here

aeneasr: I mean the file! ;)

helm/charts/hydra/templates/job-mirgation.yaml
piotrmsc: How this will block new instances? I see you do the `test -f && k delete` but this will always delete the cm?

helm/charts/hydra/files/automigrate.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# If port was given, we use it, if not, empty var is expanded to 0
+until nc -zv -w 5 $DB_URL $DB_PORT; do
+  echo "$DB_URL not yet ready"
+  sleep 5
+done
+
+hydra migrate sql -e --yes --config /etc/config/config.yaml
+echo {\"timestamp\": \"$(date "+%Y%m%d-%H%M%S")\",\"completed\": true} > /tmp/migration.json
+kubectl create configmap "${NAME_ROOT}-migration" -n "${NAMESPACE}" --from-file=/tmp/migration.json
```
piotrmsc, line 40: maybe add a retry here to do not retry everything if last CM create fails in the first try? let's assume migration did work but `kubectl` receives for example `connection refused` or `timeout`, then we will repeat the entire operation which will be unnecessary time consuming, and we may create artificial load on db unnecessary even though cli will tell us that migration was already done.

helm/charts/hydra/files/automigrate.sh
piotrmsc: but why? since this is actual script that does the work? or you want to have it inlined in the job definition (hope not ;) )

helm/charts/hydra/values.yaml
Demonsthere: In our images we have an `ory` user which has pid 100, why not use it :)

piotrmsc:
we need to test upgrade from previous version to this version as CM is part of helm release in the currently released version

helm/charts/hydra/templates/configmap.yaml
```diff
@@ -5,6 +5,10 @@ metadata:
   namespace: {{ .Release.Namespace }}
   labels:
 {{ include "hydra.labels" . | indent 4 }}
+  annotations:
+    helm.sh/hook-weight: "0"
+    helm.sh/hook: "pre-install, pre-upgrade"
+    helm.sh/hook-delete-policy: "before-hook-creation"
```
piotrmsc, line 11: this delete policy hook may be risky: if upgrade fails for some reason, helm will delete the CM already leaving hydra without config.  If someone is having rollback this could help him but if someone is not using rollback may run into trouble.

helm/charts/hydra/templates/job-mirgation.yaml
```diff
@@ -0,0 +1,71 @@
...
+      volumes:
+        - name: {{ include "hydra.name" . }}-config-volume
+          configMap:
+            name: {{ include "hydra.fullname" . }}
+        - name: {{ include "hydra.name" . }}-migration
+          configMap:
+            name: {{ include "hydra.fullname" . }}-migration
+            optional: true
+      serviceAccountName: {{ include "hydra.fullname" . }}-automigrate
+  backoffLimit: 1
```
piotrmsc, line 70:
```suggestion
  backoffLimit: 3
```

helm/charts/hydra/files/automigrate.sh
aeneasr: Maybe I don't understand what this is doing - what is the script for?

helm/charts/hydra/values.yaml
aeneasr: Ah! makes sense :)

helm/charts/hydra/files/automigrate.sh
Demonsthere: For easier development we keep the file outside (don't need to worry about yaml indent, and can run some linters on it easier), and include it into the job here: https://github.com/ory/k8s/pull/248/files#diff-9474db6f15521172343a9f42d42a172ba3143f5d9e742727d8fdb955382b3982R33

piotrmsc (approved):
👍  LGTM

helm/charts/hydra/files/automigrate.sh
aeneasr: But I don't understand what this script is for? It is creating a config map with a JSON that has a timestamp and it is pinging the database? But why do we need that? The Hydra CLI is already doing all of that for us and it is unit, functional, and e2e tested where as this script is not which will make it hard to maintain / merge patches plus it is doing something that is already solved. Hence my suggestion to not use this approach

Demonsthere: Ok I think I see your point. The idea behind the CM was to not schedule hydra pods until migration was done (the deployment would wait until the CM was created). The nc role is to wait for the db to be ready, no point in keeping the pods in a crash state until the db is up and running.

aeneasr (approved):

PR #252, opened 2021-04-22
feat: Allow configuring probes values

patryk:
## Related issue

During one of our internal deployments which was related to a fix for the runtime issue, we had an instance that the Hydra deployment was taking more time connecting to the database than usual. Unfortunately that meant it took more than preconfigured probe checks in Hydra's chart failing our deployment.

## Proposed changes

This PR allows for configuring custom probe timeouts by adding new values to the chart. The existing values are preserved as default values.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added necessary documentation within the code base (if
      appropriate).

PR #268, opened 2021-05-25
feat: allow configuring probe values for kratos & keto

JeffreyThijs:
## Related issue

Similar to #252, during deployment we noticed that the readiness check of the Kratos service took longer than the default `timeoutSeconds` parameter. However, there is no option to adapt this parameter. 

## Proposed changes

This PR allows for configuring custom probe timeouts by adding new values to the chart for Kratos and Keto, similarly as #252 did for Hydra.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/keto/templates/deployment.yaml
+++ b/helm/charts/keto/templates/deployment.yaml
@@ -50,10 +50,12 @@ spec:
             httpGet:
               path: /health/alive
               port: http-write
+            {{- toYaml .Values.deployment.livenessProbe | nindent 12 }}
           readinessProbe:
             httpGet:
               path: /health/ready
               port: http-write
+          {{- toYaml .Values.deployment.readinessProbe | nindent 12 }}
           resources:
             {{- toYaml .Values.resources | nindent 12 }}
           env:
--- a/helm/charts/keto/values.yaml
+++ b/helm/charts/keto/values.yaml
@@ -167,3 +167,14 @@ tracing:
 tolerations: []
 
 affinity: {}
+
+# Configure the probes for when the deployment is considered ready and ongoing health check
+deployment:
+  livenessProbe:
+    initialDelaySeconds: 30
+    periodSeconds: 10
+    failureThreshold: 5
+  readinessProbe:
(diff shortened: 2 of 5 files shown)


helm/charts/keto/values.yaml
```diff
@@ -164,3 +164,13 @@ tracing:
 tolerations: []
 
 affinity: {}
+
+# Configure the probes for when the deployment is considered ready and ongoing health check
+livenessProbe:
```
aeneasr, line 169: Could you move those under deployment? :)

helm/charts/keto/templates/deployment.yaml
```diff
@@ -49,10 +49,16 @@ spec:
             httpGet:
               path: /health/alive
               port: http-write
+            initialDelaySeconds: {{ .Values.livenessProbe.initialDelaySeconds }}
```
aeneasr, line 52: Move these to deployment :)

aeneasr (changes requested):
Awesome, thank you for your contribution! This looks pretty good and I have some ideas how to improve it further :)

Demonsthere:
Hi there, I'd prefer a bit different approach, using `toYaml` like we do [here](https://github.com/ory/k8s/blob/master/helm/charts/hydra-maester/templates/deployment.yaml#L43)

helm/charts/keto/values.yaml
JeffreyThijs: Ok, i'll adapt it but a deployment does not exist yet for in the keto chart although i would like to add it to be consistent with the other charts.

JeffreyThijs:
> Hi there, I'd prefer a bit different approach, using `toYaml` like we do [here](https://github.com/ory/k8s/blob/master/helm/charts/hydra-maester/templates/deployment.yaml#L43)

Looks cleaner indeed.

Demonsthere (approved):
LGTM :)

aeneasr (approved):
Awesome, thank you! 🎉 Your contribution makes Ory better :)

PR #270, opened 2021-05-27
feat: add Istio support to kratos

cortopy:
## Related issue

N/A

## Proposed changes

This PR adds opt-in configuration options to use Kratos with Istio

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

This PR doesn't introduce any changes to existing users. All the modifications are configurable and optional.

This PR adds the following:

- Service account (disabled by default) for the deployment and the migration job. I think having a dedicated service account is best practice any way, but in the case of Istio, it's useful to write [AuthorizationPolicy](https://istio.io/latest/docs/reference/config/security/authorization-policy/)
- There was a Kubernetes proposal for sidecars that got stalled. For now, any cluster using service meshes (Linkerd, istio) needs to manage jobs differently. See https://github.com/istio/istio/issues/6324. This PR makes the args and command of the migration job configurable for this. I've written an example in the `values.yaml` file for reference

### DIFF
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -36,6 +36,9 @@ spec:
       imagePullSecrets:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+    {{- if .Values.serviceAccount.enabled }}
+      serviceAccountName: {{ default (include "kratos.fullname" . ) .Values.serviceAccount.name }}
+    {{- end }}
     {{- if .Values.deployment.extraInitContainers}}
       initContainers:
         {{- if .Values.deployment.extraInitContainers }}
--- a/helm/charts/kratos/templates/job-migration.yaml
+++ b/helm/charts/kratos/templates/job-migration.yaml
@@ -24,12 +24,19 @@ spec:
         {{- toYaml . | nindent 8 }}
     {{- end }}
     spec:
+      {{- if .Values.serviceAccount.enabled }}
+      serviceAccountName: {{ default (include "kratos.fullname" . ) .Values.serviceAccount.name }}
+      {{- end }}
       containers:
       - name: {{ .Chart.Name }}-automigrate
         image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
         imagePullPolicy: {{ .Values.image.pullPolicy }}
-        command: ["kratos"]
-        args: ["migrate", "sql", "-e", "--yes"]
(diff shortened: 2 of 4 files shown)


Demonsthere:
Hello there, I have to say I have mixed feelings with such an approach. Don't you think, it would be enough to add sidecar annotations to the job rather than modify the entrypoint? Afaik, the init job shouldn't be added to the mesh.

cortopy:
What annotations do you have in mind? 

My postgresql pod is already in the mesh. Istio handles mTLS for any connection to the database, and so the job needs to be part of it too. I wouldn't like to open an unencrypted endpoint just for the job

Demonsthere:
I was thinking about adding `sidecar.istio.io/inject=false` to prohibit the pod from being added to the mesh, then you can also create an [Policy](https://istio.io/latest/docs/tasks/security/authentication/authn-policy/) to not require mTLS from this pod.

Demonsthere:
I think it is possible to configure istio to not require mTLS for a particular connection, however I see why You would not want to do to that. My proposition: instead of modyfying the entrypoint, and requiring building your own images, maybe set the /quitquitquit call as a [pod hook](https://kubernetes.io/docs/concepts/containers/container-lifecycle-hooks/) ?

cortopy:
unfortunately container lifecycle doesn't work. See someone already reported on it https://stackoverflow.com/questions/54921054/terminate-istio-sidecar-istio-proxy-for-a-kubernetes-job-cronjob

The problem is that there is one pod and two containers. Container A (the job itself) finishes, but container B (Envoy proxy) doesn't know that Container A has exited and so stays up. Kubernetes doesn't see anything wrong with this since Container B doesn't have health checks and it's not crashing, so it keeps the pod running for ever.

Container lifecycle works for all containers at once. If one stays alive, then they won't kick in

Demonsthere (approved):
Technically: this is correct, and after resolving conflicts we can merge it. 
However, I don't really like adding such a specific workaround for istio :/ but completely understand the need for it

aeneasr (commented):
Thank you for the PR! As @Demonsthere said this looks quite specific to a particular use case as it appears to require custom
built images? As such I’m not sure if this belongs into upstream. Helm render + customize would probably work better?

cortopy:
all service meshes at the moment require hacks to make it work with jobs. I thought some people may benefit from this, but I totally get that workarounds are leaking to where it shouldn't

I'll close this PR then. Thanks for considering it!

PR #279, opened 2021-06-15
feat: allow probe successThreshold & timeoutSeconds configuration

JeffreyThijs:
## Proposed changes

Allow configuration of `successThreshold` & `timeoutSeconds` for liveness and readiness probes,which I wanted to achieve with #268 but apparently forgot to configure.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/hydra/values.yaml
+++ b/helm/charts/hydra/values.yaml
@@ -236,10 +236,14 @@ deployment:
     initialDelaySeconds: 30
     periodSeconds: 10
     failureThreshold: 5
+    successThreshold: 1
+    timeoutSeconds: 1
   readinessProbe:
     initialDelaySeconds: 30
     periodSeconds: 10
     failureThreshold: 5
+    successThreshold: 1
+    timeoutSeconds: 1
 
   # https://github.com/kubernetes/kubernetes/issues/57601
   automountServiceAccountToken: false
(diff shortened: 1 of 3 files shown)


aeneasr:
Thank you! Wouldn't these default to some kubernetes system settings?

aeneasr:
Probably also doesn't make sense to have a default initial delay of 30 seconds? @Demonsthere what do you think?

Demonsthere:
The default k8s values are `timeout=1s period=10s #success=1 #failure=3`, I think the point here is to have them be configurable by values, which is now possible by installing the applications like:

```bash
cat << EOF > my-values.yaml
deployment:
  livenessProbe:
    initialDelaySeconds: 24
    periodSeconds: 7
    failureThreshold: 3
    successThreshold: 4
    timeoutSeconds: 2
EOF
helm instal .... -f my-values.yaml
```

We do not need to manually specify them in the default values. Regarding the 30s delay, it is rather but shouldn't hurt, unless someone has very strict monitoring/alerting rules, as the app won't start probes before the 30s mark

JeffreyThijs:
Oh yeah righ i forgot that you can simply specify all the probe values now since we use `{{- toYaml .Values.deployment.livenessProbe  ... }}` notation. So this PR is redundant.

aeneasr:
While the PR is being worked on I will mark it as a draft. That declutters our review backlog :)

Once you're done with your changes and would like someone to review them, mark the PR as ready and request a review from one of the maintainers.

Thank you!

JeffreyThijs:
Closing this PR due to the fact that you can all probe values with the `ToYaml` notation.

Issue #285, opened 2021-06-23
Upgrade default Kratos version to 0.6: a few challenges & options


zhming0:
**Is your feature request related to a problem? Please describe.**

Currently, there are a few problems with the Kratos Chart. 

1. Default Kratos image is still v0.5. 
2. Default Kratos image use the SQLite version. 
3. If user overrides the image tag to v0.6, the user has no way to spin up the `courier`, neither for the singleton setup nor the multi-instance setup.

Therefore, there is some work needed to be done for the v0.6 upgrade. 

**Describe the solution you'd like**

Introduce a new singleton `courier` deployment. 

There are a few viable options for this chart upgrade:

1. [Recommended] Change the default tag to use non-sqlite version `v0.6.3-alpha.1`, allow user opting out the courier deployment, and opt in the `--watch-courier` option. 
    * Pro: fits most users' workflow. 
2. Keep the current tradition, change the default tag to `v0.6.3-alpha.1-sqlite`, allow user to opt-in the courier deployment option. 
    * Pros: Backward compatible.
3. [Best long term] Reconsider the necessity of `kratos courier watch`, a DB transaction-based solution might be an alternative option to avoid the race condition of multiple couriers.
    * Pros: no need for the `courier` deployment. Simpler. Scalable.
    * Cons: depending on DB and implementation, might incur small performance penalty
    
    
I am happy to help.  Keen to know your thoughts

aeneasr:
In our system we are using a StatefulSet for the courier watcher which allows us to run it as a long-running process but also be able to up/downgrade it. We haven't ported that to the open source yet though.

Generally, I think we should probably disable `--watch-courier` in kubernetes as most people probably run it with a replication factor > 1. I think supporting SQLite still makes sense though, you could write the DB to a PVC which would be a minimalistic zero-dependency but yet still working deployment!

zhming0:
Cool, I think we are on the same page. We will make `--watch-courier` opt-in for people who wants to use SQLite. 

Meanwhile, @aeneasr do you mind approving #286 for now? It offers a quick temporary path for people (like me) wanting to deploy 0.6 to prod. 

---

Lastly, I am still a bit unsure why the courier worker has to be a singleton. I understand we use DB as a task queue, and this should allow parallel processing. Is there a particular reason we want it to be singleton? Like strictly FIFO?

aeneasr:
[...]

PR #286, opened 2021-06-23
feat: Upgrade kratos to 0.6 (quick win)

zhming0:
## Related issue

#285 

## Proposed changes

This PR implements the first cut of option 2 in #285. 
It's not a complete solution but it will give user a way to deploy v0.6 to production right now.

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

It will be so great if we make a quick release for this. I am currently stuck on upgrading v0.5 -> v0.6. 
I have done all the code change and only realize the chart isn't ready yet .

### DIFF
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -66,7 +66,7 @@ spec:
           args: [
             "serve",
             "all",
-            "--watch-courier" # Will be configurable, check out https://github.com/ory/k8s/issues/285
+            "--watch-courier", # Will be configurable, check out https://github.com/ory/k8s/issues/285
             {{- if .Values.kratos.development }}
             "--dev",
             {{- end}}
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -66,6 +66,7 @@ spec:
           args: [
             "serve",
             "all",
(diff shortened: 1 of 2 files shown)


helm/charts/kratos/templates/deployment.yaml
```diff
@@ -66,6 +66,7 @@ spec:
           args: [
             "serve",
             "all",
+            "--watch-courier", # Will be configurable, check out https://github.com/ory/k8s/issues/285
             {{- if .Values.kratos.development }}
```
Demonsthere, line 70:
```suggestion
            "all",
            {{- if .Values.kratos.config.courier.watch }}
            "--watch-courier", # Will be configurable, check out https://github.com/ory/k8s/issues/285
            {{- end }}
            {{- if .Values.kratos.development }}
```

Demonsthere (changes requested):
Hello there, thanks for the PR! 
I think that for now, we have to make this opt-in, which means setting the argument under a conditional flag (see suggestion). and not modifying the image in default values, but rather setting them during installation.

helm/charts/kratos/templates/deployment.yaml
zhming0: Thanks for the reply @Demonsthere. 

However, I have decided to close this PR in favor of a proper one #289. 

It's because I realize:

1. the changes in this PR will make user even more confused
2. More work is needed to make this chart usable. 

so, if you don't mind, please check out #289. 😃

PR #289, opened 2021-06-25
feat: Uprade Kratos 0.6:  Courier StatefulSet, new auto migrate mechanism, changing default tag to non-sqlite version

### LINKED ISSUE #285: Upgrade default Kratos version to 0.6: a few challenges & options

zhming0:
## Related issue

Close #285

Close #286 

## Proposed changes

My apology for this relatively big PR. The amount of necessary changes are beyond my expectation too. 
I won't be unhappy if you reject this change. 

As per the discussion with @aeneasr in #285. This PR introduces a few changes:

1. Upgrade default tag to v0.6.3-alpha.1 (no SQLite)
2. Introduce a courier StatefulSet.
3. Enable standalone courier by default. 
4. Allow user to opt-in the `--watch-courier` option.

Some extra changes are necessary:
1. Stop using Helm hook as the auto migration mechanism. Instead, I use initContainer to proactively wait the completion of migration job. Helm hook is a client side feature, it gets executed before everything else, including other dependencies. It quickly break things, if the migration task relies on other ConfigMaps or chart dependencies.
2. In order to get point 1 working, I had to introduce rbac to the chart so initContainer can have access to Kube API's job resources. 
3. To support point 1 and 2, I have also slightly change that `automountServiceToken` is calculated instead of set. 

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

## Further comments

I have been using the new chart on deploying my product to production.

### DIFF
--- a/helm/charts/kratos/templates/_helpers.tpl
+++ b/helm/charts/kratos/templates/_helpers.tpl
@@ -99,3 +99,20 @@ app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
 {{- end }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 {{- end -}}
+
+{{- define "kratos.courier.name" -}}
+{{- include "kratos.name" . }}-courier
+{{- end -}}
+
+{{/*
+Courier statefulset labels
+*/}}
+{{- define "kratos.courier.labels" -}}
+app.kubernetes.io/name: {{ include "kratos.courier.name" . }}
+helm.sh/chart: {{ include "kratos.chart" . }}
+app.kubernetes.io/instance: {{ .Release.Name }}
+{{- if .Chart.AppVersion }}
+app.kubernetes.io/version: {{ .Chart.AppVersion | quote }}
+{{- end }}
+app.kubernetes.io/managed-by: {{ .Release.Service }}
+{{- end -}}
--- a/helm/charts/kratos/templates/_helpers.tpl
+++ b/helm/charts/kratos/templates/_helpers.tpl
@@ -25,6 +25,14 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 
 {{/*
+Name of the Automigrate Job
+It's unique per release
+*/}}
+{{- define "kratos.automigrate-job-name" -}}
+{{ include "kratos.fullname" . }}-automigrate-{{ .Release.Revision }}
+{{- end }}
+
+{{/*
 Create a secret name which can be overridden.
 */}}
 {{- define "kratos.secretname" -}}
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -58,6 +58,13 @@ spec:
         {{- end }}
         {{- end }}
       automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
+      initContainers:
+        - name: wait-for-automigrate
+          image: "groundnuty/k8s-wait-for:v1.4"
+          imagePullPolicy: IfNotPresent
+          args:
+            - job
+            - {{ include "kratos.automigrate-job-name" . }}
       containers:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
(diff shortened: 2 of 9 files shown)


aeneasr:
@zhming0 thank you so much for working on this! We have decided to push our internal chart we are using in our prod system at console.ory.sh to this repository to keep feature parity between the two. We have included some cool gimmicks also which we're using there. Sorry for not being able to merge this PR!

aeneasr:
If you have any changes you'd like from this PR in kratos chart feel free to create another PR :)

zhming0:
hey @aeneasr no worries at all, I totally understand 😄 . Thank you for the reply. 

I can point out a few issues with the merged PR. 

1. It offers no way for opting in `--watch-courier`, so currently all SQLite users won't be able to upgrade (if they exist). 
2. As a library chart, Helm Hook is an infeasible way to run migration. It will work in your cloud env as the main chart. But quickly falls apart if Kratos is used as part of dependencies along with other charts like DB. I can elaborate more should you wish.

My PR contain fixes to both issues so I will have to stick to my own chart for the time being. 

Please let me know if you'd like to address any of these issues. I am happy to contribute.

Demonsthere:
Just my few cents:
- Jobs are problematic entity, as by their definition, they are immutable. Meaning, upgrading the chart to a version with a new image may fail, as the existing job cannot be modified. That was the reason of using hooks, to run it and delete it before the installation and upgrade phases. 
- the `--watch-couier` could be a nice additions, please feel free to add in a future PR 😉 
Nevertheless, you are right. The migrations are a problem, and we need to find a better way of handling them in the broader context of ory applications

zhming0:
> Jobs are problematic entity, as by their definition, they are immutable. Meaning, upgrading the chart to a version with a new image may fail, as the existing job cannot be modified.

Hi @Demonsthere 😃, you are right. K8s Job are immutable, and that's something we can take advantage of. 

In my PR, you can see, for every new release (in the context of `helm upgrade`), there will be a new job with a unique name generated. So upgrading won't be a issue.

aeneasr:
We experimented a lot with jobs and stateful sets and SS are the most unproblematic when upgrading and changing things as the job never completes / finishes and would need to be killed.

Regarding watch courier I think the current set up is preferable because we don’t need guards for e.g. deployment > 2 etc :)

zhming0:
@aeneasr There might be a misunderstanding here... I wasn't arguing against the StatefulSet. I was arguing against using Helm Hook. 

Let me explain a bit more.
Imagine, if a user chart has MySQL as dependency 1 and Kratos as dependency 2. The user wants the Kratos to depends on the MySQL chart. The Helm Hook mechanism will break this dependency. Helm hook will start migration job before even trying to spin up the MySQL chart. 

So, if we are on the same page, would you want me to address this problem using the solution in this PR? 

---

Regarding watch courier, I don't care SQLite use case either 😆

aeneasr:
That shouldn't be a problem because kratos migration has exponential backoff retries. So yeah, it might be not running for a bit but eventually it will apply the migrations

zhming0:
@aeneasr This would be a problem because the helm hook is designed to block until the job completes successfully and **only then** it would create other resources.

Taking one step back, if we were hoping to rely on the exponential backoff retries to make the migration work, then we might as well stop using helm hook 🙂.

I hope it make senses.

PR #296, opened 2021-07-13
feat: improve parsing of secret envs

Demonsthere:
## Related issue(s)
- Fixes https://github.com/ory/hydra/issues/2609

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/hydra/templates/_helpers.tpl
+++ b/helm/charts/hydra/templates/_helpers.tpl
@@ -81,15 +81,15 @@ Generate the name of the secret resource containing secrets
 Generate the secrets.system value
 */}}
 {{- define "hydra.secrets.system" -}}
-{{- if .Values.hydra.config.secrets.system -}}
-{{- if (kindIs "string" .Values.hydra.config.secrets.system) -}}
-{{- printf "'%s'" ( .Values.hydra.config.secrets.system | nospace | trim ) -}}
-{{- else -}}
-{{- fail "The value of .Values.hydra.config.secrets.system is incorrect. Please view values.yaml for examples" -}}
-{{- end -}}
-{{- else if .Values.demo -}}
-a-very-insecure-secret-for-checking-out-the-demo
-{{- end -}}
+  {{- if .Values.hydra.config.secrets.system -}}
+    {{- if kindIs "slice" .Values.hydra.config.secrets.system -}}
+      {{- printf "'%s'" ( mustToRawJson .Values.hydra.config.secrets.system ) -}}
+    {{- else -}}
+      {{- fail "Expected .Values.hydra.config.secrets.system to be a list of strings" -}}
+    {{- end -}}
+  {{- else if .Values.demo -}}
+'["a-very-insecure-secret-for-checking-out-the-demo"]'
+  {{- end -}}
 {{- end -}}
 
 {{/*
--- a/helm/charts/hydra/templates/_helpers.tpl
+++ b/helm/charts/hydra/templates/_helpers.tpl
@@ -81,15 +81,15 @@ Generate the name of the secret resource containing secrets
 Generate the secrets.system value
 */}}
 {{- define "hydra.secrets.system" -}}
-  {{- if .Values.hydra.config.secrets.system -}}
-    {{- $tmp := "" -}}
-    {{- range .Values.hydra.config.secrets.system -}}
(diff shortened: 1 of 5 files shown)


helm/charts/hydra/templates/_helpers.tpl
```diff
@@ -81,11 +81,15 @@ Generate the name of the secret resource containing secrets
...
 {{- define "hydra.secrets.system" -}}
-{{- if .Values.hydra.config.secrets.system -}}
-{{- .Values.hydra.config.secrets.system }}
-{{- else if .Values.demo -}}
-a-very-insecure-secret-for-checking-out-the-demo
-{{- end -}}
+  {{- if .Values.hydra.config.secrets.system -}}
+    {{- $tmp := "" -}}
+    {{- range .Values.hydra.config.secrets.system -}}
+      {{- $tmp = printf "%s,'%s'" $tmp . -}}
```
zepatrik, line 87: This will print single quotes around values? That is currently not properly parsed by hydra, see upstream PR https://github.com/ory/x/pull/373

Demonsthere: This will print the supplied secrets as `'foo bar 123 456 lorem','foo bar 123 456 lorem 1','foo bar 123 456 lorem 2','foo bar 123 456 lorem 3'`. It seems to be working aka, hydra is running and not complaining about them

zepatrik: yes it is not complaining but the secrets are now actually including the quote, therefore they are different

zepatrik: and then hydra fails when trying to use those secrets later on for older records

zepatrik: TL;DR
We have to use some csv encoding function here instead, that also properly escapes commas and double quotes.

zepatrik: Actually I just saw that we can pass JSON as env vars as well:
https://github.com/ory/x/blob/137e7b6b3a38c0af1a4b94879c5cbd9a8d406a90/configx/koanf_env.go#L44-L47

So in fact, it would be better to just output this as a JSON array of strings.

Demonsthere: Makes sense, but then we don't really need any code changes to the charts, rather a doc change to show the proper, expected syntax

zepatrik: Well we have to because the chart currently does not support secrets containing commas and/or spaces.
Or rather results in unexpected output.

zepatrik: Ahh I see so you want people to put encoded JSON themselves?

Demonsthere: yes i think it makes the most sense, currently we force a string, which can be either a simple value or stringied json array

zepatrik: OK not the best user experience though.
Why not 
```gotemplate
{{- define "hydra.secrets.system" -}}
  {{- if .Values.hydra.config.secrets.system -}}
    {{- if kindIs "[]string" .Values.hydra.config.secrets.system -}}
      {{- mustToRawJson .Values.hydra.config.secrets.system -}}
    {{- else -}}
      {{- fail "Expected .Values.hydra.config.secrets.system to be a list of strings" -}}
    {{- end -}}
  {{- else if .Values.demo -}}
  ["a-very-insecure-secret-for-checking-out-the-demo"]
  {{- end -}}
```

Demonsthere: This could also work, we just need to stringify at the end, as the result would be interpreted in bash `[ string1 string2]` and hydra doesn't like that

zepatrik: It is up to you, I just think that it would be a bad user experience to put raw JSON into the yaml, instead of automatically encoding the yaml values when rendering the template.

Demonsthere: I had to modify it slightly, but was able to make it work ;)

zepatrik (approved):
Looks good to me :wink:

Demonsthere:
@aeneasr What do you think?

aeneasr (approved):

PR #231, opened 2021-03-13
Added extraInitContainers to Kratos helm chart

junky89:
## Related issue

Implementation of the following discussion:
[https://github.com/ory/kratos/discussions/1111](https://github.com/ory/kratos/discussions/1111)

@tricky42 

## Proposed changes

The change is done because I needed to execute an init container so I could download cockroachdb credentials to connect to the database. The "extraInitContainers" is placed **before** the migration init container in the deployment template file so the migration can have the proper credentials to connect to the bd when executed. 

Maybe I should name the "extraInitContainers" "preExtraInitContainers"?

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [ ] I have added necessary documentation within the code base (if
      appropriate).

## Further comments

This is my first pull request ever, so sorry beforehand if I mistake something. 

As an example on how I used it:
```
...
deployment:
  annotations: {}
  environmentSecretsName: null
  extraEnv: []
  labels: {}
  nodeSelector: {}
  resources: {}
  tolerations: []
  tracing:
    datadog:
      enabled: false
  extraInitContainers: |
    - name: init-certs
      image: cockroachdb/cockroach-k8s-request-cert:0.4
      imagePullPolicy: IfNotPresent
      command:
      - "/bin/ash"
      - "-ecx"
      - "/request-cert -namespace=${POD_NAMESPACE} -certs-dir=/cockroach-certs -type=client -user=root -symlink-ca-from=/var/run/secrets/kubernetes.io/serviceaccount/ca.crt && chown -R 100:101 /cockroach-certs"
      env:
      - name: POD_NAMESPACE
        valueFrom:
          fieldRef:
            fieldPath: metadata.namespace
      volumeMounts:
      - name: client-certs
        mountPath: /cockroach-certs
  extraVolumes:
    - name: client-certs
  extraVolumeMounts:
    - name: client-certs
      mountPath: /cockroach-certs
...
```

PR #300, opened 2021-07-31
feat(kratos): add initContainers to kratos courier + migrations

scottcrossen:
https://github.com/ory/k8s/pull/231 added `extraInitContainers` to only the main deployment. These init-containers should probably have been added to both the migration job and courier so that mTLS can be setup correctly for these pods which also require access to the database (see the example in their PR)

I personally don't like that we're using `tpl` here instead of `toYaml` but I've kept it for consistency with the initContainer in the deployment. I can change it upon request.

Tested manually

CLA signed based on California's Labor Code Section 2870 Notice. (Employer non-assignable inventions)

## Related issue(s)
Pull request: https://github.com/ory/k8s/pull/231

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/job-migration.yaml
+++ b/helm/charts/kratos/templates/job-migration.yaml
@@ -44,6 +44,10 @@ spec:
 {{ toYaml .Values.deployment.extraVolumeMounts | indent 10 }}
          {{- end }}
       restartPolicy: Never
+      {{- if .Values.deployment.extraInitContainers }}
+      initContainers:
+{{ toYaml ( tpl .Values.deployment.extraInitContainers . ) | indent 8 }}
+      {{- end }}
       {{- if .Values.deployment.extraVolumes }}
       volumes:
 {{ toYaml .Values.deployment.extraVolumes | indent 8 }}
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -42,7 +42,7 @@ spec:
     {{- if .Values.deployment.extraInitContainers}}
       initContainers:
         {{- if .Values.deployment.extraInitContainers }}
-{{ tpl .Values.deployment.extraInitContainers . | indent 8 }}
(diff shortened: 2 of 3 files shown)


helm/charts/kratos/Chart.yaml
```diff
@@ -2,5 +2,5 @@ apiVersion: v2
 appVersion: "0.6.3-alpha.1"
 description: A ORY Kratos Helm chart for Kubernetes
 name: kratos
-version: 0.18.0
+version: 0.18.1
```
Demonsthere, line 5: Hi there, please do not modify this file, that is bumped on release ;)

helm/charts/kratos/templates/job-migration.yaml
```diff
@@ -44,6 +44,10 @@ spec:
 {{ toYaml .Values.deployment.extraVolumeMounts | indent 10 }}
          {{- end }}
       restartPolicy: Never
+      {{- if .Values.deployment.extraInitContainers }}
+      initContainers:
+{{ tpl .Values.deployment.extraInitContainers . | indent 8 }}
```
Demonsthere, line 49: I think the tpl here is used in cases you need to pass data into the init container. This would allow you to specify for example `{{ .Release.Namespace }}` as a value for some env variable in the container. But i agree, in some cases this could lead to invalid parsing. Maybe it would be best to encapsulate it like `toYaml ( tpl .Values.deployment.extraInitContainers . ) | indent 8` ?

Demonsthere (changes requested):

scottcrossen:
Sorry for the delay on this. I'll fix the PR in one week once I finally get time

helm/charts/kratos/templates/job-migration.yaml
scottcrossen: Great tip!

scottcrossen:
Updated and rebased. Thanks for being patien @Demonsthere !

Demonsthere (approved):
Lgtm!

Issue #305, opened 2021-08-03
Update the Ingress objects for all charts 


Demonsthere:
**Is your feature request related to a problem? Please describe.**

In #299 we have updated the Ingress definition to be compliant with newest version of k8s and noticed that the manifests differs between charts (for example compare [oathkeeper](https://github.com/ory/k8s/blob/master/helm/charts/oathkeeper/templates/ingress-api.yaml#L3) and [keto](https://github.com/ory/k8s/blob/master/helm/charts/keto/templates/ingress-write.yaml#L3)). 

**Describe the solution you'd like**

Update all charts to have the same Ingress definition using the current version of k8s (1.18+)

**Describe alternatives you've considered**

A clear and concise description of any alternative solutions or features you've
considered.

**Additional context**

This is a breaking change, as the charts will no longer work on older version of k8s (< 1.18). This however should not be a big issue, as that version of k8s is already EOL

- Followup to https://github.com/ory/k8s/pull/299

piotrmsc:
Once this is finished and released we need to make sure it is mentioned (dropping support for k8s <1.18) on the release itself as kind of release note

Demonsthere:
Yeah, and in the readme of each release we need to include the upgrade steps requires, so we can link them in the Release notes

PR #306, opened 2021-08-04
feat: update all ingress objects for k8s 1.18+

Demonsthere:
## Related issue(s)
- Implements https://github.com/ory/k8s/issues/305

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).


helm/charts/kratos/values.yaml
```diff
@@ -56,29 +56,31 @@ secret:
 ingress:
   admin:
     enabled: false
+    className: ""
```
piotrmsc, line 59: Will it be sufficient/work? Reading https://kubernetes.io/blog/2020/04/02/improvements-to-the-ingress-api-in-kubernetes-1.18/#setting-a-default-ingressclass tells it will work if one has annotated IngressClass CR but what if hasn't ?

piotrmsc:
And I would add in the repo root README also an info about min k8s version supported

helm/charts/keto/templates/ingress-read.yaml
```diff
@@ -1,42 +1,52 @@
 {{- if .Values.ingress.read.enabled -}}
 {{- $fullName := include "keto.fullname" . -}}
+{{- if semverCompare ">=1.19-0" .Capabilities.KubeVersion.GitVersion -}}
```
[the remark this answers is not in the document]
Demonsthere: I don't think so. Remember that in some versions, you will have both ApiVersions, and only use the newest one once the older versions has been deleted. Reacting on k8s version I think makes us more proactive, as we can react to changes faster

helm/charts/kratos/values.yaml
Demonsthere: The resulting manifests will be correct and valid from helm/k8s viewpoint. Whether they will work OOTB? I don't think so, not expect them to. Users will need to supply proper values, and they can do that, as the required config is exposed via values

helm/charts/keto/templates/ingress-read.yaml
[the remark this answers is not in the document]
piotrmsc: It might be, yes, but are you 100% certain that `.Capabilities.KubeVersion.GitVersion` will always be in the format `1.19-0` ?  Cause I'm not 100% sure and in such cases this if won't give meaningful result. 
You are right, we do not provide a way for user to explicitly tell which version to use so we can either : 
- change it to provide some additional flag - then user can impact decision with overrides
- go with apiversion check for newest stable version first -  still does not allow to pick version but ensures stable API first over alpha/beta

helm/charts/kratos/values.yaml
piotrmsc: General comment: charts should work in minimal setup OOTB after doing `helm install`, if there are any specific **required** values to be provided in order to work or other prerequisites, They have to be described as well as the behaviour. I would put that desc in chart readme.

piotrmsc:
Hmm, do we have already a task to move to GH actions or we keep on with CircleCI?
And I think it will be good in the future (not in this PR ofc) to add k3d and helm install steps and test if installation from helm process work.

Demonsthere:
> Hmm, do we have already a task to move to GH actions or we keep on with CircleCI?
> And I think it will be good in the future (not in this PR ofc) to add k3d and helm install steps and test if installation from helm process work.

Dude :D We have that already, but with kind: https://app.circleci.com/pipelines/github/ory/k8s/1010/workflows/a3d5ef9d-78c4-4bf9-92a5-25791746838b/jobs/2093 

Regarding GH, no, and to be honest, can't really see the benefit of moving to them right now

aeneasr (approved):
Can't really say much about the k8s contents, but everything else looks nice!

helm/charts/keto/templates/ingress-read.yaml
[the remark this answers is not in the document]
Demonsthere: After some digging: the GitVersion is taken from the git release tag, which uses semVer, so the format is like ` GitVersion:"v1.21.1"`. However the function [semverCompare](https://helm.sh/docs/chart_template_guide/function_list/#semvercompare) is flexible enough for it to work

helm/charts/kratos/values.yaml
Demonsthere: Yes, this is still valid. Please remember that ingress objects are optional, and disabled by default. Enabling them requires the user to provide additional input/configuration. The checks I've added in this PR ensure that:
- charts work OOTB - they can be installed, have to validation errors, and the pods are in a running state
- ingress objects are valid and configurable

If we'd like to check if the ingress objects work, we would need to install an ingress controller and modify the hosts file to check connectivity. But this is imho a lot of overhead and testing if k8s features works, rather then ours

piotrmsc: What I meant is add an info for Ingress what is required to make it work OOTB if one enables it

piotrmsc (approved):
General OK

helm/charts/keto/templates/ingress-read.yaml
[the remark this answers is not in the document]
piotrmsc: I was reflecting more to forks of K8s as kubernetes providers do not use 1:1 OS k8s ;-)

PR #310, opened 2021-08-15
fix(kratos): correct initContainer formatting

scottcrossen:
https://github.com/ory/k8s/pull/300 Added support for init-containers, but the way they were implemented led helm to interpret them as strings instead of as yaml. This was an error of communication between author and reviewer.

Example of how this is currently parsed:

```
initContainers:
|
- command: ...
```

I'm not sure why exactly the bar gets added but it does

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -42,7 +42,7 @@ spec:
     {{- if .Values.deployment.extraInitContainers}}
       initContainers:
         {{- if .Values.deployment.extraInitContainers }}
-{{ toYaml ( tpl .Values.deployment.extraInitContainers . ) | indent 8 }}
+{{ tpl .Values.deployment.extraInitContainers . | indent 8 }}
         {{- end }}
     {{- end }}
       volumes:
--- a/helm/charts/kratos/templates/job-migration.yaml
+++ b/helm/charts/kratos/templates/job-migration.yaml
@@ -46,7 +46,7 @@ spec:
       restartPolicy: Never
       {{- if .Values.deployment.extraInitContainers }}
       initContainers:
-{{ toYaml ( tpl .Values.deployment.extraInitContainers . ) | indent 8 }}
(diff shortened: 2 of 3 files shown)


scottcrossen:
@Demonsthere  Sorry for the miscommunication on this. I should have tested it after I made reviewer-suggested changes.

helm/charts/kratos/templates/deployment.yaml
```diff
@@ -42,7 +42,7 @@ spec:
     {{- if .Values.deployment.extraInitContainers}}
       initContainers:
         {{- if .Values.deployment.extraInitContainers }}
-{{ toYaml ( tpl .Values.deployment.extraInitContainers . ) | indent 8 }}
+{{ tpl .Values.deployment.extraInitContainers . | indent 8 }}
```
scottcrossen, line 45: I think we should leave this the way it was before https://github.com/ory/k8s/pull/300 to avoid changing behavior too much. Again, I don't like using `tpl` because that makes the values.yaml file look weird, but I prefer consistency.

Demonsthere: The idea of using `tpl` is a common in some charts, that's why we use it. For example [here](https://github.com/helm/charts/blob/master/incubator/elasticsearch/ci/extrainitcontainers-values.yaml)

Demonsthere:
@scottcrossen my bad, i assumed the CI will test the change, but we don't have any init containers (even dummy ones) in the CI. Is this something you could add in your PR? An entry in the [values files](https://github.com/ory/k8s/blob/master/.circleci/values/kratos.yaml) that will and an dummy init container (printing hello world), just so it will be templated and verified

Demonsthere (approved):

scottcrossen:
@Demonsthere Thanks for the merge! I was about to say "yes I can add some tests" but it looks like you already got that. Thanks for that too!

Issue #318, opened 2021-09-01
Charts Should Be Compatible With HPA


samcgardner:
**Is your feature request related to a problem? Please describe.**

It's currently impossible to use the deployments in these charts alongside the HPA in a GitOps setup, since you are forced to specify the `replicas` value. This means that any changes the HPA makes will be immediately reversed by the GitOps agent. 

EDIT: 

Minor addendum - Keto doesn't suffer from this issue since it ships its own HPA and correctly removes the `replicas` field from the deployment template if it is configured.

**Describe the solution you'd like**

Change the deployment templates like so to make the `replicas` value optional:

```yaml
{{ if  .values.replicaCount }}
  replicas: {{ .Values.replicaCount }}
{{ end }}
```

**Describe alternatives you've considered**

I don't see another plausible solution to this problem.

**Additional context**

I'm happy to PR this if you're willing to review.

piotrmsc:
I reproduced the original issue easily on local env with : 
```
k3d create cluster test

argocd app create hydra --repo https://github.com/ory/k8s.git --path helm/charts/hydra  --dest-server https://kubernetes.default.svc --dest-namespace default --helm-set hydra.config.secrets.system={SOME_SECRET} --helm-set hydra.config.dsn=memory --helm-set hydra.config.urls.self.issuer=https://my-hydra/ --helm-set hydra.config.urls.login=https://my-idp/login --helm-set maester.enabled=false --helm-set image.tag=v1.10.5-sqlite --helm-set deployment.autoscaling.enabled=true --helm-set deployment.autoscaling.minReplicas=2 --helm-set deployment.autoscaling.maxReplicas=3 --helm-set deployment.autoscaling.targetCPUUtilizationPercentage=2 --helm-set deployment.resources.requests.cpu=10m --sync-policy auto --self-heal

```

PR #319, opened 2021-09-03
fix(hpa): remove problematic explicit setting of replica counts in deployments

samcgardner:
This change fixes https://github.com/ory/k8s/issues/318 

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/hpa.yaml
+++ b/helm/charts/kratos/templates/hpa.yaml
@@ -1,5 +1,5 @@
 {{- if .Values.autoscaling.enabled }}
-apiVersion: autoscaling/v2beta1
+apiVersion: autoscaling/v2beta2
 kind: HorizontalPodAutoscaler
 metadata:
   name: {{ include "kratos.fullname" . }}
--- a/helm/charts/kratos/templates/hpa.yaml
+++ b/helm/charts/kratos/templates/hpa.yaml
@@ -1,0 +1,28 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "kratos.fullname" . }}
+  labels:
+    {{- include "kratos.labels" . | nindent 4 }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ include "kratos.fullname" . }}
+  minReplicas: {{ .Values.autoscaling.minReplicas }}
+  maxReplicas: {{ .Values.autoscaling.maxReplicas }}
+  metrics:
+  {{- if .Values.autoscaling.targetCPUUtilizationPercentage }}
+    - type: Resource
+      resource:
+        name: cpu
+        targetAverageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
+  {{- end }}
+  {{- if .Values.autoscaling.targetMemoryUtilizationPercentage }}
+    - type: Resource
+      resource:
+        name: memory
+        targetAverageUtilization: {{ .Values.autoscaling.targetMemoryUtilizationPercentage }}
+  {{- end }}
+{{- end }}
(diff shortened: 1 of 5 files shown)


helm/charts/kratos/templates/hpa.yaml
```diff
@@ -0,0 +1,28 @@
+{{- if .Values.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
```
piotrmsc, line 2: nit: In hydra we already use beta2
```suggestion
apiVersion: autoscaling/v2beta2
```

helm/charts/kratos/values.yaml
```diff
@@ -267,6 +267,14 @@ securityContext:
   allowPrivilegeEscalation: false
   privileged: false
 
+# Horizontal pod autoscaling configuration
+autoscaling:
```
piotrmsc, line 271: In the issue you've mentioned kratos is not impacted however I see you did the same change as in hydra?
Also with hydra by default we give HPA with max to 3 replicas (we don't have demo profiles etc) so maybe let's limit a bit 100 replicas ;-)

piotrmsc:
Thank you for your contribution :) Does it work for you in your git ops env? I'm asking because 
with this change, after templating charts with autoscaling enabled deployment won't have replicas set, HPA will set it but then if this deployment is watched by the controller it will notice that `replicas` property has been and will try to remove it as part of reverting to the original state.  Or it ignores new fields added to the spec?

What tools do you use ?:) I assume you have gitOps controller + some helm controller to deploy stuff picked up by git controller?

helm/charts/kratos/values.yaml
samcgardner: That seems reasonable. I think Keto uses 100 as its maximum, so you might want to look at changing that as well.

samcgardner:
We use ArgoCD and its default Helm controller. I'm not entirely sure _why_ this works. Now you mention it, it does seem logical for the controller to try and prune the extra field. Perhaps ArgoCD specifically ignores added replica fields for this exact reason?

piotrmsc:
yeah, exactly, I had the same thought that actually it is still the same case. I will take a look a bit deeper to fully understand the behavior

piotrmsc:
for future reference to reproduce the issue easily locally do: 
```
k3d create cluster test

argocd app create hydra --repo https://github.com/ory/k8s.git --path helm/charts/hydra  --dest-server https://kubernetes.default.svc --dest-namespace default --helm-set hydra.config.secrets.system={SOME_SECRET} --helm-set hydra.config.dsn=memory --helm-set hydra.config.urls.self.issuer=https://my-hydra/ --helm-set hydra.config.urls.login=https://my-idp/login --helm-set maester.enabled=false --helm-set image.tag=v1.10.5-sqlite --helm-set deployment.autoscaling.enabled=true --helm-set deployment.autoscaling.minReplicas=2 --helm-set deployment.autoscaling.maxReplicas=3 --helm-set deployment.autoscaling.targetCPUUtilizationPercentage=2 --helm-set deployment.resources.requests.cpu=10m --sync-policy auto --self-heal

```

piotrmsc:
@samcgardner I run into this issue https://github.com/argoproj/argo-cd/issues/1603 where exactly same behavior is described.

However, I think it should be possible to define which fields should be ingored from sync decision in argocd like it's mentioned here https://argoproj.github.io/argo-cd/user-guide/diffing/#application-level-configuration

piotrmsc:
but looks like ignoring replicas this way will not always work https://github.com/argoproj/argo-cd/issues/2913

samcgardner:
Yes, I think this approach would probably work as well. I'm not sure the developer ergonomics are as good, or that it would transfer well to other projects like Flux, so personally I'd prefer to offer this solution in the chart as well since it's fairly standard and harmless

piotrmsc:
I want to understand the behavior more precisely as it's pretty much the same for me in both scenarios TBH.
On ArgoCD slack I received confirmation that ignoring field while diffing is a way to go but no explanation why it behaves differently when replicas are not set in the manifest and then added by HPA.

piotrmsc (approved):
But in general LGTM! Thank you for the contribution 🎉

PR #322, opened 2021-09-17
feat(hydra-maester): Allow configuration of the sync-period parameter.

d0x2f:
This PR allows configuring the sync-period command line parameter of hydra-maester.

https://github.com/ory/hydra-maester/blob/028c3df4c2b8455f4f0fba27e11507490b140590/main.go#L62

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/hydra-maester/README.md
+++ b/helm/charts/hydra-maester/README.md
@@ -25,7 +25,6 @@ A Helm chart for Kubernetes
 | deployment.tolerations | list | `[]` | k8s toleration |
 | enabledNamespaces | list | `[]` | The Controller have CREATE and READ access to all Secrets in the namespaces listed below. |
 | forwardedProto | string | `nil` |  |
-| syncPeriod | string | `10h` | The minimum frequency at which watched resources are reconciled |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
 | image.repository | string | `"oryd/hydra-maester"` |  |
 | image.tag | string | `"v0.0.23"` |  |
--- a/helm/charts/hydra-maester/README.md
+++ b/helm/charts/hydra-maester/README.md
@@ -25,6 +25,7 @@ A Helm chart for Kubernetes
 | deployment.tolerations | list | `[]` | k8s toleration |
 | enabledNamespaces | list | `[]` | The Controller have CREATE and READ access to all Secrets in the namespaces listed below. |
 | forwardedProto | string | `nil` |  |
+| syncPeriod | string | `10h` | The minimum frequency at which watched resources are reconciled |
 | image.pullPolicy | string | `"IfNotPresent"` |  |
 | image.repository | string | `"oryd/hydra-maester"` |  |
 | image.tag | string | `"v0.0.23"` |  |
--- a/helm/charts/hydra-maester/templates/deployment.yaml
+++ b/helm/charts/hydra-maester/templates/deployment.yaml
@@ -43,8 +43,8 @@ spec:
             {{- if .Values.singleNamespaceMode }}
             - --namespace={{ .Release.Namespace }}
             {{- end }}
-            {{- if .Values.syncPeriod }}
-            - --sync-period={{ .Values.syncPeriod }}
+            {{- if .Values.deployment.args.syncPeriod }}
+            - --sync-period={{ .Values.deployment.args.syncPeriod }}
             {{- end }}
           resources:
             {{- toYaml .Values.deployment.resources | nindent 12 }}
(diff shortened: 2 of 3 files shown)


helm/charts/hydra-maester/README.md
```diff
@@ -25,6 +25,7 @@ A Helm chart for Kubernetes
 | deployment.tolerations | list | `[]` | k8s toleration |
 | enabledNamespaces | list | `[]` | The Controller have CREATE and READ access to all Secrets in the namespaces listed below. |
 | forwardedProto | string | `nil` |  |
+| syncPeriod | string | `10h` | The minimum frequency at which watched resources are reconciled |
```
Demonsthere, line 28: This file is generated by https://github.com/norwoodj/helm-docs, just wanted to make sure you didn't modify it manually ;)

helm/charts/hydra-maester/values.yaml
```diff
@@ -26,7 +26,9 @@ adminService:
   port:
 
 forwardedProto:
-    
+
+syncPeriod: 10h
```
Demonsthere, line 30: i'd prefer to move the value to either `deployment.args.syncPeriod` or to be more in line with our other charts `maester.config.syncPeriod`

helm/charts/hydra-maester/templates/deployment.yaml
```diff
@@ -43,6 +43,9 @@ spec:
             {{- if .Values.singleNamespaceMode }}
             - --namespace={{ .Release.Namespace }}
             {{- end }}
+            {{- if .Values.syncPeriod }}
+            - --sync-period={{ .Values.syncPeriod }}
```
Demonsthere, line 47: I have one issue here. The change isn't yet released, so the deployment will fail due to an unexpected input flag, as there is a default value, so this will always be set. 
Please either remove the redundant if, or the default value :)

Demonsthere (changes requested):
Hi there! Thanks for the PR, i left a few small comments :)

helm/charts/hydra-maester/README.md
d0x2f: ah, I did! I'll add the comment to the Values.yml file and leave the readme alone then.

helm/charts/hydra-maester/templates/deployment.yaml
d0x2f: Good point, I'll remove the default value then.

This way users won't need to update their Values.yml file when they upgrade the chart.

Demonsthere: in that case please do it like this:

```yaml
# The minimum frequency at which watched resources are reconciled
syncPeriod: ""
# syncPeriod: 10h
```

helm/charts/hydra-maester/README.md
d0x2f: Is it intended that I run the tool to generate the readme and include it in this PR?
When I run helm-docs I get a very sparse readme missing a lot of the text.

helm/charts/hydra-maester/values.yaml
d0x2f: I've moved it to deployment.args.syncPeriod.

helm/charts/hydra-maester/templates/deployment.yaml
d0x2f: Done as you said.

helm/charts/hydra-maester/README.md
Demonsthere: I see, it seems in some release of the tool, something in the description changed, right now it requires syntax like 
```
# -- The Controller have CREATE and READ access to all Secrets in the namespaces listed below.
enabledNamespaces: []
```
seems the `--` is required. I will create a issue for fixing that globally :)

helm/charts/hydra-maester/templates/deployment.yaml
```diff
@@ -43,6 +43,9 @@ spec:
             {{- if .Values.singleNamespaceMode }}
             - --namespace={{ .Release.Namespace }}
             {{- end }}
+            {{- if .Values.syncPeriod }}
```
Demonsthere, line 46:
```suggestion
            {{- if .Values.deployment.args.syncPeriod }}
```

helm/charts/hydra-maester/templates/deployment.yaml
Demonsthere:
```suggestion
            - --sync-period={{ .Values.deployment.args.syncPeriod }}
```

Demonsthere (commented):
Seems something is missing ;)

d0x2f:
Woops! suggestions applied.

Demonsthere (approved):
LGTM :)

PR #331, opened 2021-09-24
chore: Update maester versions

Demonsthere:
## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/.circleci/values/oathkeeper-maester.yaml
+++ b/.circleci/values/oathkeeper-maester.yaml
@@ -1,0 +1,2 @@
+---
+# singleNamespaceMode: true
--- a/helm/charts/oathkeeper-maester/templates/deployment.yaml
+++ b/helm/charts/oathkeeper-maester/templates/deployment.yaml
@@ -52,12 +52,10 @@ spec:
             - --rulesFileName={{ .Values.rulesFileName }}
             {{- end }}
           env:
+            {{- if .Values.singleNamespaceMode }}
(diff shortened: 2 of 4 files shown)


helm/charts/oathkeeper-maester/templates/deployment.yaml
```diff
@@ -51,14 +51,20 @@ spec:
             {{- if .Values.rulesFileName }}
             - --rulesFileName={{ .Values.rulesFileName }}
             {{- end }}
+          env:
+            - name: NAMESPACE
+              {{- if .Values.singleNamespaceMode }}
+              value: {{ .Release.Namespace }}
+              {{- else }}
+              value: ""
+              {{- end }}
```
piotrmsc, line 60: We don't need to duplicate os.getenv behaviour here, let's set NS env if it's enabled by the flag

.circleci/values/oathkeeper-maester.yaml
```diff
@@ -0,0 +1,2 @@
+---
+# singleNamespaceMode: true
```
piotrmsc, line 2: why do we have it here commented out?

Demonsthere: we don't want it to be set by default, but having this here makes debugging easier, as it is enough to uncomment it and use the file :)

piotrmsc (approved):

PR #346, opened 2021-10-18
fix(HPA): fix kratos HPA by using the correct autoscaling v1 cpu utilization config

### LINKED ISSUE #340: Kratos HPA doesn't work when activating it
**Describe the bug**

When we add the autoscaling config and enable it we get an exception thrown during the deployment.

> unknown field "targetAverageUtilization" in io.k8s.api.autoscaling.v2beta2.ResourceMetricSource, ValidationError(HorizontalPodAutoscaler.spec.metrics[0].resource): missing required field "target" in io.k8s.api.autoscaling.v2beta2.ResourceMetricSource

**To Reproduce**

1. Add the following configuration
 
 ```
  autoscaling:
    enabled: true
    minReplicas: 1
    maxReplicas: 3
    targetCPUUtilizationPercentage: 80
  ```
2. Do a helm install 

**Expected behavior**

I'm expecting the chart to install everything without an issue.

**Environment**

- Version: 0.19.5
- Environment: Minikube

**Additional context**
The issue is because the template is using [autoscaling/v2beta2](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v2beta2/) which doesn't contain the `targetAverageUtilization`. That field is only present in the [autoscaling/v1](https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/horizontal-pod-autoscaler-v1/).

If we want to use autoscaling/v2beta2, we should be switching to the new format that defines the target.

```
{{- if .Values.autoscaling.targetCPUUtilizationPercentage }}
  metrics:
    - type: Resource
      resource:
        name: cpu
        target:
          type: Utilization
          averageUtilization: {{ .Values.autoscaling.targetCPUUtilizationPercentage }}
  {{- end }}
```

abdoofathy95:
## Related issue(s)

This Change fixes #340 and adds useAutoscalingApiV2beta2 flag to support both autoscaling/v1 & autoscaling/v2beta2

## Checklist

      and signed the CLA.
      the changes.
      works.
      appropriate).

### DIFF
--- a/.circleci/values/kratos.yaml
+++ b/.circleci/values/kratos.yaml
@@ -1,6 +1,6 @@
 ---
 autoscaling:
-  enabled: true
+  enabled: false
 kratos:
   autoMigrate: true
   identitySchemas:


aeneasr:
Unfortunately, for some reason, the CircleCI tests are not running. Do you maybe follow your/a fork of this repository on CircleCI? If so, you need to unsubscribe / unwatch from that CircleCI project. Then, make another push to your branch using:

```
$ git commit --amend --allow-empty
$ git push --force
```

That should get the CI running! Thank you :)

Demonsthere:
Ok, so the problem is because the current release still has the bug described in https://github.com/ory/k8s/issues/340 and cannot be installed with 

```yaml
autoscaling:
  enabled: true
```

Installing current kratos with autoscaling disabled and upgrading to this branch with enabled works. Could you now disable autoscaling in the overrides? We will merge it, and change after a bugfix release :)

Issue #105, opened 2020-01-07
Hydra / Maester: Make the install of custom CRDs optional and rely on Helm 3 convention for CRDs installation


PoulpiFr:
**Is your feature request related to a problem? Please describe.**
My current dev workflow (Skaffold + Helm 3) is doing a lot of `helm upgrade` when iterating. Hydra Maester chart is taking a long time to be upgraded, due to the usage of a Job to install CRDs.

As there are now standardized way of installing CRDs with Helm 3 that would be faster / more standard, would it be possible to offer a way to rely on them?

**Describe the solution you'd like**
- Make the installation of CRDs through a job at each `helm upgrade` **optional** with a new value parameter (that can default to true)
- Move the CRD file into a `/crds` directory so that Helm 3 can install them at the first chart install only.

**Describe alternatives you've considered**
I see no other solution. Change my dev workflow maybe? 

However, skaffold and Helm seems to be popular choices.

**Additional context**
https://developer.ibm.com/blogs/kubernetes-helm-3/#simplified-custom-resource-definition-crd-support
https://github.com/helm/helm/issues/5871
https://github.com/helm/helm/pull/6243

-> Please note that this would prevent this chart from updating the CRDs without changing its apiVersion from what I understand. While the current behavior is doing it.

If you're OK with the approach, I can do a PR.

aeneasr:
We have unfortunately not yet worked with Helm 3 so I have personally very little input on that. What's important for us is to keep Helm 2 compatibility.

Regarding moving the CRD (previously this was done with hooks iirc) - how would that affect CRD updates?

However, I think this really boils down to another issue, which is this one: https://github.com/ory/k8s/issues/103

It appears that this flag has no effect which might actually be what you're looking for?

aeneasr:
Yeah, the `/crd` directory only works in Helm 3, I don't think it would be a good idea to make these charts incompatible with helm 2.

Demonsthere:
Closed by https://github.com/ory/k8s/pull/235 :)

PR #214, opened 2021-01-07
feat(hydra-maester): make hydra-maester pods labels customizable

MatthieuFin:
## Related issue

none

## Proposed changes

This PR allows the user of the chart to customize the labels of hydra-maester pod and job crd-init. This feature permit us to match global network policies which allow kubernetes api access, which is not available by default.

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security
      vulnerability. If this pull request addresses a security vulnerability, I
      confirm that I got green light (please contact
      [security@ory.sh](mailto:security@ory.sh)) from the maintainers to push
      the changes.
- [ ] I have added tests that prove my fix is effective or that my feature
      works.
- [x] I have added necessary documentation within the code base (if
      appropriate).

## Further comments
- CLA form respond with 404 error
- I didn't open an issue about this and don't find an existing. I choose to purpose feature insteed of issue.
- I use this chart on an "on premise" k8s cluster and hope merge it to community version.

PR #351, opened 2021-10-26
fix: Use actual repository url for maester dependency

remotezygote:
The relative file path in use currently breaks outside of straight helm usage, for example when using this chart with `garden`. This also adds `*.txt` to the `.helmignore` files in each of `hydra` and `oathkeeper` as those files were being "built" by helm when testing locally.

## Related issue(s)

https://github.com/ory/k8s/issues/105#issuecomment-952267328

## Checklist

      and signed the CLA.
      the changes.
      works. (I don't know how to test this)
      appropriate).


Demonsthere:
I took the liberty of doing some housekeeping for this PR ;) It still seams the CLA is unhappy as https://github.com/ory/k8s/pull/351/commits/4637bf3c3fa78425b79d092b2365425ebff58e02 seems to be signed with a different email. Could you try following https://github.com/ory/k8s/pull/214#issuecomment-756615018 to fix that?

helm/charts/hydra/Chart.yaml
```diff
@@ -26,4 +26,4 @@ dependencies:
     version: 0.20.1
     condition: maester.enabled
     alias: hydra-maester
-    repository: file://../hydra-maester
+    repository: https://k8s.ory.sh/helm/charts
```
Demonsthere, line 29: I like this approach, clean and simple :) 
One thing thou, I would like to keep the second, file url as a comment. It is helpful to have it for local development (f.e if you want to modify the maester, and test that for hydra/oathkeeper)

remotezygote:
```suggestion
    # repository: file://../hydra-maester
    repository: https://k8s.ory.sh/helm/charts
```

helm/charts/oathkeeper/Chart.yaml
```diff
@@ -27,4 +27,4 @@ dependencies:
     version: 0.20.1
     condition: maester.enabled
     alias: oathkeeper-maester
-    repository: file://../oathkeeper-maester
+    repository: https://k8s.ory.sh/helm/charts
```
Demonsthere, line 30:
```suggestion
    # repository: file://../oathkeeper-maester
    repository: https://k8s.ory.sh/helm/charts
```

Demonsthere (approved):

remotezygote:
Hm, the release where this is supposed to be seems to have changed these lines at some point?

https://github.com/ory/k8s/blob/v0.20.2/helm/charts/hydra/Chart.yaml#L29-L30

Demonsthere:
Hi @remotezygote yes, I had to revert it for now, as it is a breaking change to the current release process (didn't noticed that beforehand :/). As we release all charts at once, during the release 0.20.2 of oathkeeper, helm tries to download the release 0.20.2 of maester, which is not yet available. This can be seen [here](https://app.circleci.com/pipelines/github/ory/k8s/1221/workflows/a9bb1ffa-60de-4aac-aeed-4b1ad5f4b271/jobs/2752/parallel-runs/0/steps/0-102). 
I'm sorry. but we cannot use the repo url, unless we modify the release pipeline first, to for example work in sequential and upload the charts one after another, or separate the process to be per-chart 😞

remotezygote:
Yeah, was trying to use it with `garden` and it is unworkable in its current state. My only solution is to clone the charts locally in my repo which I don't want to do for the obvious maintenance reasons. I decided to go another route, so won't be pushing this forward more. I would highly recommend doing this, though - people who are using `garden` or similar tools are gonna get pretty frustrated with it and likely give up (kinda what I did).

remotezygote:
Oh! And disabling maester does nothing to disable installing it, which is the issue here. It can't install if its directory structure doesn't match yours. That's not how it's supposed to work. You cannot install the `hydra` or `oathkeeper` charts _at all_ in `garden` because of this.

PR #375, opened 2021-11-21
fix: Enable sqa-opt-out CLI flag

tristanls:
The only remaining method of disabling SQA is a CLI flag. See https://github.com/ory/kratos/issues/1959. This pull request allows configuring the `--sqa-opt-out` CLI flag in the kratos helm chart.

https://github.com/ory/kratos/issues/1959

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -98,7 +98,7 @@ spec:
             {{- if .Values.kratos.development }}
             - --dev
             {{- end }}
-            {{- if .Values.kratos.config.sqa-opt-out }}
+            {{- if .Values.kratos.sqaOptOut }}
             - --sqa-opt-out
             {{- end }}
             - --config
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -90,13 +90,16 @@ spec:
         - name: {{ .Chart.Name }}
           image: {{ include "kratos.image" . }}
           imagePullPolicy: {{ include "kratos.imagePullPolicy" . }}
-          command: 
+          command:
             - kratos
           args:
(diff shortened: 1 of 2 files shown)


aeneasr:
https://github.com/ory/kratos/issues/1959 is a buf which needs to be fixed in ory/x :)

helm/charts/kratos/templates/deployment.yaml
```diff
@@ -90,14 +90,17 @@ spec:
...
-          command: 
+          command:
             - kratos
           args:
             - serve
             - all
             {{- if .Values.kratos.development }}
             - --dev
             {{- end }}
+            {{- if .Values.kratos.sqaOptOut }}
```
Demonsthere, line 101: I'd prefer to tackle this in a more generic way, like we use for extraEnvs, volumes etc:

```yaml
            {{- if .Values.deployment.extraEnv }}
{{ toYaml .Values.deployment.extraEnv | indent 12 }}
            {{- end }}
```

```suggestion
            {{- if .Values.deployment.extraArgs }}
```

And

Demonsthere (changes requested):
Hello there!
Thanks for the PR :) I like the idea with passing extra arguments, but would prefer a more generic approach. Please see my suggestion

helm/charts/kratos/templates/deployment.yaml
tristanls: This is a better improvement than what I proposed @Demonsthere and would be helpful in general 👍 .

Demonsthere: Yeah, I think it would be beneficial and more generic. If you want to update your PR and add it, please reopen this one or create a new one :)

Issue #388, opened 2022-01-16
Add different labels to public and admin services of Kratos


lusu007:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

I'm am trying to integrate Ory Kratos with the [prometheus-stack](https://github.com/prometheus-community/helm-charts) in my development cluster. The [prometheus-stack](https://github.com/prometheus-community/helm-charts) can discover services that it should monitor by creating a custom resource called ServiceMonitor. To select the correct service a [label selector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/) is used. Currently, a selection of the admin or public service via labels is not possible, because both have exactly the same label set.

At the moment  [prometheus-stack](https://github.com/prometheus-community/helm-charts) tries to monitor both services at once. But this isn't possible because the metrics endpoint is only available on the Kratos admin service. Therefore Prometheus is throwing a 404 error.
![image](https://user-images.githubusercontent.com/10341994/149754646-b325e38d-5f35-4642-8dcb-c7c6f3238e69.png)

### Describe your ideal solution

I would suggest using one of the Kubernetes [recommended labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/) `app.kubernetes.io/component`.

The label could be set to `app.kubernetes.io/component=admin` on the Kratos admin service and to `app.kubernetes.io/component=public` on the Kratos public service.

### Workarounds or alternatives

A possible workaround is to set the labels extra labels via the `values.yaml` file. However, I think that the two services should already be differentiable by label by default.

### Version

v0.21.5

PR #389, opened 2022-01-17
feat: Add component label to kratos svcs

lusu007:
Implements #388 

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/service-admin.yaml
+++ b/helm/charts/kratos/templates/service-admin.yaml
@@ -5,6 +5,7 @@ metadata:
   name: {{ include "kratos.fullname" . }}-admin
   namespace: {{ .Release.Namespace }}
   labels:
+    app.kubernetes.io/component: admin
 {{ include "kratos.labels" . | indent 4 }}
   {{- with .Values.service.admin.annotations }}
   annotations:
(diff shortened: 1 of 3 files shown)


Demonsthere (commented):
Hi there and thanks for the PR :) 
The current solution is good, but what do you think about making this a bit more generic? 
Instead of adding the labels hardcoded like this, i'd prefer to use 
`service.admin.labels`, like we have with annotations

```yaml
{{- with .Values.service.admin.labels }}
    {{- toYaml . | nindent 4 }}
{{- end }}
```

lusu007:
Thank you for your comment 😊

The possibility to add custom labels via `values.yaml` should be added in any case. But shouldn't the component label always be set, since it helps to differentiate the two services? 😅

Demonsthere:
> The possibility to add custom labels via `values.yaml` should be added in any case. But shouldn't the component label always be set, since it helps to differentiate the two services? 😅

Yes, of course, i meant to set those labels as default, and then append any user provided ones to them. Sorry! 😅

Demonsthere (approved):
Great, thanks!

Issue #402, opened 2022-02-27
Resources declaration level isn't consistent


thomasboni:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

In the `deployment` template, the resource specification is retrieved from root:

```yaml
          resources:
            {{- toYaml .Values.resources | nindent 12 }}
```

But in values file, it's located under `deployment`:

```yaml
deployment:

  [...]

  resources: {}
  #  We usually recommend not to specify default resources and to leave this as a conscious
  #  choice for the user. This also increases chances charts run on environments with little
  #  resources, such as Minikube. If you do want to specify resources, uncomment the following
  #  lines, adjust them as necessary, and remove the curly braces after 'resources:'.
  #  limits:
  #    cpu: 100m
  #    memory: 128Mi
  #  requests:
  #    cpu: 100m
  #  memory: 128Mi
```

This isn't consistent and can lead to configuration errors for users. I see two solutions:

- Update the template to search for `resources` under `deployment`: more consistent but will be a **breaking changes** for user currently uses `resrouces` at root level in their values files
- Update the default values files by putting `resources` at root level: no breaking changes for users. I'm going to do this update in a PR, let me know if it's ok for you 

### Reproducing the bug

- Use the default values file
- Update `resources` section to declare some `limits` / `requests` 
- Deploy the chart
- Resources isn't' t set in `deployment`

### Version

chart v0.21.5

Demonsthere:
Hello there, thanks for raising the issue! 
In my opinion this looks like an oversight when we updated kratos charts for the new release, and added courier to it. In general we prefer the `deployment.resources` methods, as some charts, like kratos have either multiple running resources (like a job, deployment, statefulset), and we'd like to leave the possibility to specify resources per object. 
I will create a followup issue to cleanup the resource options globally :)

PR #403, opened 2022-02-27
fix: move resources declaration at root level

thomasboni:

FIx https://github.com/ory/k8s/issues/402

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

No breaking changes


Demonsthere:
Hello there, as described in the issue, i would prefer to fix the resource declaration in the deployment, as we want to split deployment, from statefulset/job resource declaration :).

thomasboni:
Hello, I've updated my PR following your remark. But I don't know what to do about the fact that this commit introduces a breaking change. 

Should I note this somewhere in a changelog file ? Or maybe in your conventional commit, I can use a prefix that specify a breaking change ?  I found nothing about it in your contributing guidelines

helm/charts/kratos/templates/deployment.yaml
```diff
@@ -187,7 +187,7 @@ spec:
           readinessProbe:
             {{- toYaml .Values.deployment.readinessProbe | nindent 12 }}
           resources:
-            {{- toYaml .Values.resources | nindent 12 }}
+            {{- toYaml .Values.deployment.resources | nindent 12 }}
```
Demonsthere, line 190: Hi again, sorry for the delay but I needed a moment to think about this. 
You are right, right now this is a change, I would not call it breaking, but it will require user input to change the chart values. 
However, we can fix it in a more smooth way:

```
if .Values.resources
  use them
else if .Values.deployment.resources
  use them
else {}
```

Could you update your PR with such a model?

Demonsthere (changes requested):

helm/charts/kratos/templates/deployment.yaml
thomasboni: Hi, ok, I see. The idea is to support `resources` both at `root` level or at `deployment` level. 

Is it something that you want to support in long term or you plan to deprecate the specification at `root` level and remove it someday ?

thomasboni: In my point of view, if we have something in `.Values.deployment.resources` it must have precedence on `.Values.resources`. What do you think ?

thomasboni:
Updated following @Demonsthere remark. I still need to test it

helm/charts/kratos/templates/deployment.yaml
Demonsthere: I think it would make more sense to deprecate the root level further on. The idea of root level resources was ok when we had only 1 application/deployment in the charts. Since now we have included jobs and others, i think it may be confusing as to what those options relate to

thomasboni: Ok, that what I understood 👍‍

Demonsthere:
Hey man, I think the pull bot overwrote your changes 😅

piotrmsc:
We have an interesting PR right now :D PR without any files changed 😂 [bot action](https://github.com/ory/k8s/compare/d310722f5184beb5ac4ea50e33f72b265a097bfd..c5b0ab76847e5bda5571a7f05f144c5d298b4db5) messed it up and there conflicts atm, overwriting your changes

thomasboni:
New PR opened here: https://github.com/ory/k8s/pull/408

PR #412, opened 2022-03-10
feat: Upgrade kratos chart to releaser 0.9.0

aeneasr:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/README.md
+++ b/helm/charts/kratos/README.md
@@ -19,10 +19,10 @@ A ORY Kratos Helm chart for Kubernetes
 | deployment.extraVolumeMounts | list | `[]` |  |
 | deployment.extraVolumes | list | `[]` | If you want to mount external volume For example, mount a secret containing Certificate root CA to verify database TLS connection. |
 | deployment.labels | object | `{}` |  |
-| deployment.livenessProbe | object | `{"failureThreshold":5,"httpGet":{"path":"/health/alive","port":"http-admin"},"initialDelaySeconds":30,"periodSeconds":10}` | Configure the probes for when the deployment is considered ready and ongoing health check |
+| deployment.livenessProbe | object | `{"failureThreshold":5,"httpGet":{"path":"/admin/health/alive","port":"http-admin"},"initialDelaySeconds":30,"periodSeconds":10}` | Configure the probes for when the deployment is considered ready and ongoing health check |
 | deployment.nodeSelector | object | `{}` | Node labels for pod assignment. |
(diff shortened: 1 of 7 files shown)


helm/charts/kratos/README.md
```diff
@@ -19,10 +19,10 @@ A ORY Kratos Helm chart for Kubernetes
 | deployment.extraVolumeMounts | list | `[]` |  |
 | deployment.extraVolumes | list | `[]` | If you want to mount external volume For example, mount a secret containing Certificate root CA to verify database TLS connection. |
 | deployment.labels | object | `{}` |  |
-| deployment.livenessProbe | object | `{"failureThreshold":5,"httpGet":{"path":"/health/alive","port":"http-admin"},"initialDelaySeconds":30,"periodSeconds":10}` | Configure the probes for when the deployment is considered ready and ongoing health check |
+| deployment.livenessProbe | object | `{"failureThreshold":5,"httpGet":{"path":"/admin/health/alive","port":"http-admin"},"initialDelaySeconds":30,"periodSeconds":10}` | Configure the probes for when the deployment is considered ready and ongoing health check |
```
Demonsthere, line 22: This file is autogenerated, no need to modify it in the PR ;)

Demonsthere (approved):

helm/charts/kratos/README.md
aeneasr: Ok, I won't next time :)

aeneasr:
Regarding release, we need to figure out how to ship this. We need this change in cloud relatively soon (today/tomorrow), but the version this targets is not yet released in the open source and its also incompatible with the older version of Ory Kratos.

Demonsthere:
If this is a breaking change, that is connected with the 0.9 release, we can add a function to check the version of the image from tag, and do if 0.9 then x, else fallback to 0.8

aeneasr:
Oh, that would be nice! Would this work for cloud also? If so, could you support me with that? :)

Alternatively I think we can also keep the old chart, as there is a 302 redirect from the old `/health/alive` to `/admin/health/alive` endpoints. So if 302 is ok for the health checker, it shouldn't be an issue I think.

Demonsthere:
According to https://github.com/kubernetes/kubernetes/issues/73172 anything between 200 and 400 is accepted :)

aeneasr:
Ok, in that case we can wait to merge this until 0.9 is out!

aeneasr:
@Demonsthere 0.9 is out

Issue #413, opened 2022-03-11
Add checksum annotation for kratos.yaml configmap 


nritholtz:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

I am currently using the kratos helm chart via argoCD for deployment.

At first I thought my changes made to the kratos.yaml via helm values weren't being propagated, since I didn't see any new deployments. Afterwards I realized that kratos itself is using a watcher on the mounted config file, but the values that I changed weren't being deployed since [serve](https://github.com/ory/kratos/blob/93a1ae4fe98487a0bca00d2afdc5e7b07c0e1c46/driver/config/config.go#L314) section of the config was immutable.

So I had to do my own "kubectl rollout" which applied annotations that argoCD saw as "out of sync" due to the new annotation that wasn't in my desired state. I know there is a concept of the [watcher](https://github.com/ory/k8s/blob/065c25d91094437132e028a3f05219d10e25e266/helm/charts/kratos/files/watch.sh#L5) included, however I'm afraid this will also cause an "out of sync" issue for argoCD, and also it seems a little superfluous from my point of view to have a watcher on the file, if I know during deployment time that something has changed.

### Describe your ideal solution

I'm proposing adding a new annotation `checksum/kratos-config`, which would be applied as follows:
```
        checksum/kratos-config: {{ include (print $.Template.BasePath "/configmap-config.yaml") . | sha256sum }}
```
as per [recommendation](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) from helm.

This would be added in the `spec.template.annotations` of the [kratos deployment template](https://github.com/ory/k8s/blob/180994d/helm/charts/kratos/templates/deployment.yaml#L38) right before user defined annotations.

[...]

PR #415, opened 2022-03-15
feat: add support for checksum annotations

nritholtz:
Introduces two new boolean values `secret.hashSumEnabled` and `configMap.hashSumEnabled`, both true by default, which will auto create/managed checksum annotations on `spec.template.metadata.annotations` for relevant pod resources. This will automatically mark those pods to be recreated after the changes.

https://github.com/ory/k8s/issues/413

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

## Further comments

I did a deep-dive into this while creating this PR, and I noticed an issue. The secrets in a few charts have auto-generated secret values by [default](https://github.com/ory/k8s/blob/master/helm/charts/kratos/templates/secrets.yaml#L17). Since I am using `include` instead of `lookup` it means every time the annotations are rendered by the helm template, it would see it as a change/different value for every upgrade (and even a different value across different references in same chart, e.g. kratos and courier). Maybe there is an alternative approach, but I wondering whether this implementation is sufficient.

In addition, I needed to do some changes to helpers where `.Values.....secrets` are an empty dict to make the reference a [safe access for null](https://stackoverflow.com/a/68807258), since it was causing issues during my development.

### DIFF
--- a/helm/charts/hydra/templates/_helpers.tpl
+++ b/helm/charts/hydra/templates/_helpers.tpl
@@ -81,7 +81,7 @@ Generate the name of the secret resource containing secrets
 Generate the secrets.system value
 */}}
 {{- define "hydra.secrets.system" -}}
-  {{- if .Values.hydra.config.secrets.system -}}
+  {{- if (.Values.hydra.config.secrets).system -}}
     {{- if kindIs "slice" .Values.hydra.config.secrets.system -}}
       {{- if gt (len .Values.hydra.config.secrets.system) 1 -}}
         "{{- join "\",\"" .Values.hydra.config.secrets.system -}}"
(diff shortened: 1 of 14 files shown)


bodymindarts:
Thank you for working on this. Just wanted to flag that I ran into exactly the same issue (config not causing a pod restart) and look forward to this being rolled out.

Demonsthere:
Hi there, I see that Kratos is failing on the PR. This might be due to the last changes related to kratos 0.9, could you rebase to the current master so the CI can run again :)?

Demonsthere:
Hmm, I see the CI failed again with `error=map[message:dsn value requested an unknown driver] service_name=Ory Kratos service_version=v0.9.0-alpha.2`, looks like the DSN value is somehow being corrupted here? Would need to run it locally and check what is happening with in the context of this changes

nritholtz:
@Demonsthere I believe I fixed the issue in the tests with my latest changes. 

https://github.com/ory/k8s/pull/415/files#diff-feeebee2e39b1cef7f251ce97c0928f8c540e4e6e6ce556a4cf767054280abfeR108 is now using `omit` and `deepCopy` instead of `unset`, since `unset` was affecting the `.Values.config` object outside the scope of the helper function. This ultimately caused the `.Values.config.dsn` to be empty for the secret when being rendered.

Demonsthere (approved):
Great job, thanks! :tada:

Issue #353, opened 2021-10-27
Don't enforce DSN in secret


toonvanstrijp:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Right now a lot of ory products use a DSN to connect with the database. This is fine, however it's causing configuration issues when you want to use a existing secret that only contains the username and password for example.

### Describe your ideal solution

The ideal solution would be to modify the helm chart so that it doesn't enforce the DSN. This way we could set it manually and have more control over how this DSN is provided.

### Workarounds or alternatives

The current workaround involves creating the DSN secret beforehand or with a initContainer or job.

### Version

v0.20.0

MOZGIII:
The most flexible way is to simply allow omitting the DSN, and letting the user to pass it via `extraEnv`

PR #416, opened 2022-03-17
fix: Do not pass DSN if it's empty

MOZGIII:

https://github.com/ory/k8s/issues/353, more specifically https://github.com/ory/k8s/issues/353#issuecomment-1071032974

Ultimately, the goal here is allow passing a custom DSN env value that's not constrained by this chart in its form/source.

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
- ~~I have added tests that prove my fix is effective or that my feature
      works.~~ existing tests are sufficient
      appropriate).

### DIFF
--- a/helm/charts/hydra/templates/_helpers.tpl
+++ b/helm/charts/hydra/templates/_helpers.tpl
@@ -111,8 +111,9 @@ Generate the secrets.cookie value
 Generate the configmap data, redacting secrets
 */}}
 {{- define "hydra.configmap" -}}
-{{- $config := unset .Values.hydra.config "dsn" -}}
-{{- $config := unset $config "secrets" -}}
+{{- $config := .Values.hydra.config | deepCopy -}}
+{{- $_ := unset $config "dsn" -}}
+{{- $_ := unset $config "secrets" -}}
 {{- toYaml $config -}}
 {{- end -}}
 
--- a/helm/charts/hydra/templates/deployment.yaml
+++ b/helm/charts/hydra/templates/deployment.yaml
@@ -133,11 +133,14 @@ spec:
             - name: URLS_SELF_ISSUER
               value: {{ $issuer | quote }}
             {{- end }}
+            {{- $dsn := include "hydra.dsn" . -}}
+            {{- if $dsn }}
             - name: DSN
               valueFrom:
                 secretKeyRef:
                   name: {{ include "hydra.secretname" . }}
                   key: dsn
+            {{- end }}
             - name: SECRETS_SYSTEM
               valueFrom:
                 secretKeyRef:
--- a/helm/charts/hydra/templates/job-migration.yaml
+++ b/helm/charts/hydra/templates/job-migration.yaml
@@ -40,11 +40,14 @@ spec:
         command: ["hydra"]
         args: ["migrate", "sql", "-e", "--yes"]
         env:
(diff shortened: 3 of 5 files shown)


MOZGIII:
We actually require it for integration with postgres operator.

```yaml
      extraEnv:
        - name: DB_USER
          valueFrom:
            secretKeyRef:
              name: hydra.hydra-db.credentials.postgresql.acid.zalan.do
              key: username
        - name: DB_PASSWORD
          valueFrom:
            secretKeyRef:
              name: hydra.hydra-db.credentials.postgresql.acid.zalan.do
              key: password
        - name: DSN
          value: "postgres://$(DB_USER):$(DB_PASSWORD)@hydra-db:5432/hydra"
```

As you can see, no value leak here, and secrets are used.

Demonsthere:
I see, and this is in cases where you don't want to/can't preload a custom secret with those values and have to provide them during installation?

```yaml
secret:
  # -- switch to false to prevent creating the secret
  enabled: false
  # -- Provide custom name of existing secret, or custom name of secret to be created
  nameOverride: "my-preexisting-secret"
```

MOZGIII:
Well, disabling the secret entirely might not be a good idea, since other values are kept there (system and cookie secrets). But it would make sense to, maybe, separate the secrets for the system/cookie from the DB credentials, and allow omitting the DB secret if the DSN is crafted manually like in our case.

MOZGIII:
Added a small fix for incorrect use of `unset` instruction at configmap

Demonsthere:
I was thinking about this problem a bit, what do you think of adding a helper function to all charts:

```yaml
{{/*
Load Envs from Secrets
*/}}
{{- define "foo.envsFromSecrets" -}}
  {{- range $sub, $envs := .Values.envsFromSecrets }}
    {{- range $it, $array := $envs }}
- name: {{ $array.key | upper }}
  valueFrom:
    secretKeyRef:
      name: {{ $array.secret }}
      key: {{ $array.key }}
    {{- end }}
  {{- end }}
{{- end }}
---
envsFromSecrets:
  db:
    - key: MY_SECRET_KEY
       secret: my-secret
```

Am I to understand that in your use case, you receive the DB credentials from the cluster, and have to construct the DSN on runtime?

MOZGIII:
> Am I to understand that in your use case, you receive the DB credentials from the cluster, and have to construct the DSN on runtime?

Yes, that is correct



> to all charts

I don't want to touch anything that I can't test in my environment

Demonsthere:
Hi there,
As much as I understand your problem here, it is a breaking change, not for the code perspective, but the contract we have with the users, in which we enforce the DSN to be supplied to us. 
Making the applications aware, or construct the DSN on the fly is an interesting feature, but it may break the current contract, that is why i am hesitant on the issue. 
To be honest, I think that in your case a valid approach might be the use of an init container, which would gather the data you need from the secrets/configs and either construct the DSN secret, or inject it to the main container. 

By all charts I meant, that if we can agree on a strategy and successfully implement it in one application, i'd like to export the solution to other charts as well

roko99:
Hello
Could you please take a look on https://github.com/ory/k8s/pull/509 as it could be acceptable here?

Demonsthere:
Closing in favour of https://github.com/ory/k8s/pull/509

PR #419, opened 2022-03-22
feat: Add support for overriding kratos deployment entrypoint (to allow sourcing DB credentials dynamically)

eh-steve:
Update kratos helm chart to allow overriding default entrypoint to support exporting environment variables using dynamic secrets.

The immediate use case is for getting a dynamic DB connection string issued by Hashicorp Vault, instead of a static kubernetes secret. 

## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -93,14 +93,14 @@ spec:
           image: {{ include "kratos.image" . }}
           imagePullPolicy: {{ include "kratos.imagePullPolicy" . }}
           command:
-          {{ if not (empty .Values.deployment.overrideArgs.command) }}
-{{ toYaml .Values.deployment.overrideArgs.command | indent 12 }}
+          {{ if .Values.deployment.customCommand }}
+{{ tpl .Values.deployment.customCommand . | indent 12 }}
           {{ else }}
             - kratos
           {{ end }}
           args:
-          {{ if not (empty .Values.deployment.overrideArgs.args) }}
-{{ toYaml .Values.deployment.overrideArgs.args | indent 12 }}
+          {{ if .Values.deployment.customArgs }}
+{{ tpl .Values.deployment.customArgs . | indent 12 }}
           {{ else }}
             - serve
             - all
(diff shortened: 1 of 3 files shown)


helm/charts/kratos/templates/deployment.yaml
```diff
@@ -93,8 +93,15 @@ spec:
           image: {{ include "kratos.image" . }}
           imagePullPolicy: {{ include "kratos.imagePullPolicy" . }}
           command:
+          {{ if not (empty .Values.deployment.overrideArgs.command) }}
```
Demonsthere, line 96: The check is redundant, you should be able to use just `if .Values.deployment.overrideArgs.command`

helm/charts/kratos/values.yaml
```diff
@@ -185,6 +185,13 @@ deployment:
   # - --sqa-opt-out
   extraArgs: []
 
+  # Set these to override the entrypoint of the deployment container
+  # (e.g. to source dynamic secrets or export environment dynamic variables)
+  overrideArgs:
+    command: []
```
Demonsthere, line 191: Not sure if an array is a good idea here. For more complex scenarios, passing a multiline string like 
```yaml
command: |
  - some very complex
  - stuff that requires {{ .Values.other.stuff }}
```

and then calling `toYaml | tpl` might be better

Demonsthere (changes requested):
Hi there, thanks for the PR!

I see the benefit of such an option, but have some remarks. 
- For consistency, I'd prefer to include it in all charts, and not only one
- In charts that have several deployment objects (deployment/statefulset/jobs) it would be smart to have it per type
- I think a better naming convention would be `customComand` and `customArgs`

helm/charts/kratos/values.yaml
Demonsthere: Please refer to the [Kartos chart](https://github.com/ory/k8s/blob/master/helm/charts/kratos/templates/deployment.yaml#L47-L51) how we handle similar parsing issues and how to [use it in values](https://github.com/ory/k8s/blob/master/.circleci/values/kratos.yaml#L83-L92)

adamstrawson:
Has there been any further progress on this? Looking to include the same behaviour in some other charts, and waiting for this to be completed so I can follow the same method for consistency.

eh-steve:
Sorry about that, we ended up implementing a few more changes in another working branch but forgot to bring them into this one - will update

Demonsthere:
Closing due to https://github.com/ory/k8s/pull/504 implementing similar feature set :)

Issue #421, opened 2022-03-23
Allow disabling securityContext


fraxachun:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [X] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

In some environments (e.g. OpenShift) you might want not to set the securityContext. Due to this behaviour of helm (https://github.com/helm/helm/issues/9136) it is not possible to unset the securityContext via own values. Also workarounds like setting the securityContext to false don't work.

### Describe your ideal solution

Add a value deployment.securityContext.enabled which allows to omit setting the securityContext. This is how the securityContext could be applied:

```
          {{- if .Values.deployment.securityContext.enabled }}
          securityContext: {{- omit .Values.deployment.securityContext "enabled" | toYaml | nindent 10 }}
          {{- end }}
```

### Workarounds or alternatives

I couldn't find any alternative to the above suggestion.

### Version

0.22.2

fraxachun:
See https://github.com/ory/k8s/pull/422

Demonsthere:
Closing as #422 was merged

PR #422, opened 2022-03-23
feat: Allow disabling securityContext

fraxachun:
Allows setting `securityContext.enabled: false` which omits the securityContext-setting for the containers. In some environment (e.g. OpenShift) you don't want to set them (e.g. they are set automatically).

https://github.com/ory/k8s/issues/421

## Checklist

- [X] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md)
      and signed the CLA.
- [X] I have referenced an issue containing the design document if my change
      introduces a new feature.
- [X] I confirm that this pull request does not address a security
      the changes.
      works.
- [X] I have added necessary documentation within the code base (if
      appropriate).

### DIFF
--- a/helm/charts/hydra-maester/templates/deployment.yaml
+++ b/helm/charts/hydra-maester/templates/deployment.yaml
@@ -54,8 +54,10 @@ spec:
             {{- toYaml .Values.deployment.resources | nindent 12 }}
           terminationMessagePath: /dev/termination-log
           terminationMessagePolicy: File
+          {{- if .Values.deployment.securityContext }}
           securityContext:
             {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+          {{- end }}
       serviceAccountName: {{ include "hydra-maester.fullname" . }}-account
       automountServiceAccountToken: {{ .Values.deployment.automountServiceAccountToken }}
       {{- if .Values.priorityClassName }}
--- a/helm/charts/hydra/templates/deployment.yaml
+++ b/helm/charts/hydra/templates/deployment.yaml
@@ -50,8 +50,10 @@ spec:
       containers:
         {{- if .Values.watcher.enabled }}
         - name: watcher
+          {{- if .Values.deployment.securityContext }}
           securityContext:
             {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+          {{- end }}
           image: {{ .Values.watcher.image }}
           command:
             - /bin/bash
@@ -153,8 +155,10 @@ spec:
             {{- end }}
           resources:
             {{- toYaml .Values.deployment.resources | nindent 12 }}
+          {{- if .Values.deployment.securityContext }}
           securityContext:
             {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+          {{- end }}
           lifecycle:
             {{- toYaml .Values.deployment.lifecycle | nindent 12 }}
         {{- if .Values.deployment.extraContainers }}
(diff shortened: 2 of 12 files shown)


Demonsthere:
Hi there!
I think this is a bit of a overkill 😅  in your values files you can just specify `securityContext: {}` to be an empty object, which should override the default values

fraxachun:
In my experience (helm v3.8.1) settting `securityContext: {}` does not work because helm is merging it with the default values. `securityContext: null` also does not work (https://github.com/helm/helm/issues/9136). I also tried `securityContext: false`, but this leads to complaining helm about the datatype and the value is rendered.

Demonsthere:
I see, and I assume you're using the charts in an umbrella chart which has the rest as sub-charts, and setting `null` does not propagate for you?

We can make `securityContext: false` work, but need to modify your PR and include `if` clauses like
```yaml
          {{- if .Values.securityContext }}
          securityContext:
            {{- toYaml .Values.securityContext | nindent 12 }}
          {{- end }}
```

helm/charts/hydra-maester/README.md
```diff
@@ -19,7 +19,7 @@ A Helm chart for Kubernetes
 | deployment.nodeSelector | object | `{}` | Node labels for pod assignment. |
 | deployment.podLabels | object | `{}` |  |
 | deployment.resources | object | `{}` |  |
-| deployment.securityContext | object | `{"allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"privileged":false,"readOnlyRootFilesystem":true,"runAsNonRoot":true,"runAsUser":1000}` | Default security context |
+| deployment.securityContext | object | `{"enabled": true",allowPrivilegeEscalation":false,"capabilities":{"drop":["ALL"]},"privileged":false,"readOnlyRootFilesystem":true,"runAsNonRoot":true,"runAsUser":1000}` | Default security context |
```
Demonsthere, line 22: Those files are auto-generated, please do not modify them by hand :)

helm/charts/hydra-maester/templates/deployment.yaml
```diff
@@ -54,8 +54,10 @@ spec:
             {{- toYaml .Values.deployment.resources | nindent 12 }}
           terminationMessagePath: /dev/termination-log
           terminationMessagePolicy: File
+          {{- if .Values.deployment.securityContext.enabled }}
           securityContext:
-            {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+            {{- omit .Values.deployment.securityContext "enabled" | toYaml | nindent 12 }}
```
Demonsthere, line 59:
```suggestion
            {{- toYaml .Values.deployment.securityContext | nindent 12 }}
```

Demonsthere: Please see the PR discussion

fraxachun:
Technically setting `securityContext: false` would work but helm complains about wrong datatypes:
```
~/dev/k8s master$ helm template test helm/charts/hydra --set deployment.securityContext=false
coalesce.go:220: warning: cannot overwrite table with non table for hydra.deployment.securityContext (map[allowPrivilegeEscalation:false capabilities:map[drop:[ALL]] privileged:false readOnlyRootFilesystem:true runAsNonRoot:true runAsUser:100])
coalesce.go:220: warning: cannot overwrite table with non table for hydra.deployment.securityContext (map[allowPrivilegeEscalation:false capabilities:map[drop:[ALL]] privileged:false readOnlyRootFilesystem:true runAsNonRoot:true runAsUser:100])
```

I would prefer the suggested solution. However, if this behaviour is acceptable I can update my PR with the `securityContext: false` solution.

Demonsthere:
Yes, please update the PR :) I think that the `securityContext: false` is a good middle point solution.

fraxachun:
@Demonsthere I force-pushed the suggested changes.

Demonsthere (approved):
Great, thanks! :)

PR #438, opened 2022-05-11
fix: deployment template smtp secret ref

iutx:
## Checklist

      and signed the CLA.
      introduces a new feature.
      the changes.
      works.
      appropriate).

### DIFF
--- a/helm/charts/kratos/templates/deployment.yaml
+++ b/helm/charts/kratos/templates/deployment.yaml
@@ -147,11 +147,13 @@ spec:
                   name: {{ include "kratos.secretname" . }}
                   key: secretsCipher
                   optional: true
+            {{- if .Values.kratos.config.courier.smtp.connection_uri }}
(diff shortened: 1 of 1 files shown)


Demonsthere (approved):
Good catch, thanks! 😃

dan2kx:
is this change needed? it means when doing an upgrade it will fail to add that secret ref and the pod will fail, unless a placeholder is used? Which happened to me...

Demonsthere:
@dan2kx I'm a bit confused by your comments, by adding the guard we should not enforce the env if it is not set in the config. Before that change the env was always required

dan2kx:
The secrets are created when the helm chart is first deployed and not on upgrade, so I would suggest this be moved to the secret chart if necessary? 

since this change, when using this chart to upgrade, I have to add a fake secret to the values file for the container to use the secret that is already in the configured in kubernetes

PR #451, opened 2022-07-04
feat: allow setting auto migration as job or initcontainer in kratos

DenisPalnitsky:
This PR changes the way Kratos runs migration in a helm chart. 

## Before
Kratos is using Job with `pre-install, pre-upgrade` to run auto migration. The problem with this approach is that when Kratos is deployed first time as a part of the umbrella helm chart that includes DB, it fails to automigrate because DB helm chart is not deployed when auto-migrate job runs. 
Here is an example to reproduce the issue: https://github.com/DenisPalnitsky/kratos-umbrella-chart

## After 
The proposed solution is to run migration as an Init container. This guarantees that DB is available when Kratos runs migration. This changes the cadence of running migration job from `once per deployment or upgrade` to `every time Kratos pod starts` however it does not affect startup performance that much. Migrations are versioned so init container finishes in seconds when DB is already migrated:
```
➜ kubectl logs kratos-847f4db75d-wbsb5 -c kratos-automigrate
time=2022-07-04T13:40:31Z level=info msg=No tracer configured - skipping tracing setup audience=application service_name=Ory Kratos service_version=v0.9.0-alpha.2
Successfully applied SQL migrations!
```

# Breaking changes 
- This patch removes kratos-automigrate job. Please ensure that you don't rely on its existence in your deployments.
- `extraInitContainers` configuration changed from string to yaml

before :
```
  extraInitContainers:  |
    - name: init-myservice
      image: busybox:1.28
      command: ['sh', '-c', "until nslookup myservice.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; do echo waiting for myservice; sleep 2; done"]
```
after:
```
 extraInitContainers:  
  - name: init-myservice
    image: busybox:1.28
    command: ['sh', '-c', "until nslookup myservice.$(cat /var/run/secrets/kubernetes.io/serviceaccount/namespace).svc.cluster.local; do echo waiting for myservice; sleep 2; done"]
```

## Checklist


DenisPalnitsky:
Please let me know if it's something you will consider to merge and I'll fix test/validate CI pipelines

Demonsthere:
Hi there! 
We actually started with a initContainer, this however came with its own set of limitations: the migration process can take a lot of time, which can cause k8s to terminate the pod in the middle of a migration, they don't have a retry/backoff or ttl mechanisms in place. 

I don't think replacing the automigration entirely is a good solution, but would happily see a possibility of picking the migration type, for example:

```yaml
automigration:
  enabled: true
  # possible values: job, initContainer, when empty defaults to job
  type: job
  job:
    job specific parameters
  initcontainer:
    init container specific parameters
```

This would allow keeping the custom initcontainers and granting the users the possibility of choosing a preferred type :) 

Is it something you'd consider reworking in this PR, or should i describe it in an issue?

DenisPalnitsky:
@Demonsthere 
That makes sense. Let me do it in this PR.

DenisPalnitsky:
@Demonsthere 
Please check it out. 

Is there any reason why `extraContainers` are declared as string in [values](https://github.com/ory/k8s/blob/a3d795290ebbad5ebbfa7986ad08bc9d17e503bc/.circleci/values/kratos.yaml#L88)?
Would it make sense to convert them to yaml as well, like [here](https://github.com/DenisPalnitsky/k8s/blob/46b5de31101ccea5702f032829f8a2f08cfedc7c/.circleci/values/kratos.yaml#L84) ?

Demonsthere:
Hi there! Great works, will try to review it as soon as i can. Regarding your question: we declared it as string, as this allows the field to be consumed as input for the `tpl` function. This allows the usage of previously defined helm values within the custom container

helm/charts/kratos/Chart.yaml
```diff
@@ -3,5 +3,5 @@ appVersion: v0.9.0-alpha.2
 description: A ORY Kratos Helm chart for Kubernetes
 name: kratos
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-kratos.svg
-version: 0.23.3
+version: 0.23.4
```
Demonsthere, line 6: This is autogenerated, please do not modify :)

helm/charts/kratos/README.md
```diff
@@ -64,7 +64,8 @@ A ORY Kratos Helm chart for Kubernetes
 | job.serviceAccount.name | string | `""` | The name of the service account to use. If not set and create is true, a name is generated using the fullname template |
 | job.shareProcessNamespace | bool | `false` | Set sharing process namespace |
 | job.spec.backoffLimit | int | `10` | Set job back off limit |
-| kratos.autoMigrate | bool | `false` | Enable the initialization job. Required to work with a DB |
+| kratos.automigration.enabled | bool | `false` | Enable the automigration. Required to work with a DB |
```
Demonsthere, line 67: This is autogenerated from comments in the values.yaml file, please do not modify manually :)

helm/charts/kratos/templates/job-migration.yaml
```diff
@@ -69,12 +69,12 @@ spec:
       restartPolicy: Never
       {{- if .Values.deployment.extraInitContainers }}
       initContainers:
-{{ tpl .Values.deployment.extraInitContainers . | indent 8 }}
+{{ toYaml .Values.deployment.extraInitContainers | indent 8 }}
```
Demonsthere, line 72: Please see my comment regarding the use of `tpl` ;)

helm/charts/kratos/templates/statefulset-mail.yaml
```diff
@@ -114,11 +113,11 @@ spec:
           {{- end }}
         {{- if $extraContainers }}
 {{ tpl $extraContainers . | indent 8 }}
-        {{- end }}
-      {{- if $extraInitContainers }}
+        {{- end }}
```
Demonsthere, line 116:
```suggestion
        {{- end }}
```

helm/charts/kratos/templates/statefulset-mail.yaml
```diff
@@ -114,11 +113,11 @@ spec:
...
         {{- if $extraContainers }}
 {{ tpl $extraContainers . | indent 8 }}
-        {{- end }}
-      {{- if $extraInitContainers }}
+        {{- end }}      
+        {{- if .Values.deployment.extraInitContainers }}
       initContainers:
-{{ tpl $extraInitContainers . | indent 8 }}
-      {{- end }}
+{{ toYaml .Values.deployment.extraInitContainers | indent 8 }}
```
Demonsthere, line 119:
```suggestion
{{- toYaml .Values.deployment.extraInitContainers | nindent 8 }}
```

helm/charts/kratos/values.yaml
```diff
@@ -105,7 +105,15 @@ ingress:
...
-  autoMigrate: false
+
+  # -- Enables database migration
+  automigration:
+    enabled: false
+    # -- Configure the way to execute database migration. Possible values: job, initContainer
+    # When set to job, the migration will be executed as a job on release or upgrade.
+    # When set to initContainer, the migration will be executed when kratos pod is created
+    # Defaults to job  
+    type: job
```
Demonsthere, line 116: To consider: we might want to add a check if the provided value is initContainer or job, and fail the installation if not. Otherwise the installation will proceed with without migration, which may result in some false-positive/false-negative errors. Imho it would be cleaner to fail in such a case

Demonsthere (changes requested):
Good job :) Adding some comments and minor improvements

helm/charts/kratos/templates/statefulset-mail.yaml
```diff
@@ -115,10 +114,10 @@ spec:
         {{- if $extraContainers }}
 {{ tpl $extraContainers . | indent 8 }}
         {{- end }}
-      {{- if $extraInitContainers }}
+        {{- if .Values.deployment.extraInitContainers }}
       initContainers:
-{{ tpl $extraInitContainers . | indent 8 }}
-      {{- end }}
+{{- tpl .Values.deployment.extraInitContainers . | nindent 8 }}
```
Demonsthere, line 119: This change will remove the possibility of specifying statefulset only init containers, please restore the ternary operation :)

Demonsthere (changes requested):
I really like how it turned out :) One little change and we can merge! I think we can make a followup issue and propagate the change to other charts as well, to have consistent operations

Demonsthere (approved):
Nice one!

PR #459, opened 2022-07-13
feat: add horizontal pod autoscaler to oathkeeper helm chart

PeteMac88:
Add horizontal pod autoscaler to oathkeeper helm chart. Oathkeeper can be scaled based on memory and (or) cpu utilization.

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.

### DIFF
--- a/helm/charts/oathkeeper/templates/hpa-controller.yaml
+++ b/helm/charts/oathkeeper/templates/hpa-controller.yaml
@@ -1,4 +1,4 @@
-{{- if and (eq .Values.global.ory.oathkeeper.maester.mode "controller") .Values.deployment.autoscaling.enabled }}
+{{- if .Values.deployment.autoscaling.enabled }}
 apiVersion: autoscaling/v2beta1
 kind: HorizontalPodAutoscaler
 metadata:
--- a/helm/charts/oathkeeper/templates/hpa-controller.yaml
+++ b/helm/charts/oathkeeper/templates/hpa-controller.yaml
@@ -1,0 +1,31 @@
+{{- if and (eq .Values.global.ory.oathkeeper.maester.mode "controller") .Values.deployment.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "oathkeeper.fullname" . }}
+  {{- if .Release.Namespace }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+  labels:
+    {{- include "oathkeeper.labels" . | nindent 4 }}
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: {{ include "oathkeeper.fullname" . }}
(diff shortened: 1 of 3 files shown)


helm/charts/oathkeeper/templates/hpa-controller.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if and (eq .Values.global.ory.oathkeeper.maester.mode "controller") .Values.deployment.autoscaling.enabled }}
+apiVersion: autoscaling/v2beta1
+kind: HorizontalPodAutoscaler
+metadata:
+  name: {{ include "oathkeeper.fullname" . }}
+  {{- if .Release.Namespace }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+  labels:
+{{ include "oathkeeper.labels" . | indent 4 }}
```
Demonsthere, line 10:
```suggestion
{{- include "oathkeeper.labels" . | nindent 4 }}
```

helm/charts/oathkeeper/values.yaml
```diff
@@ -286,6 +286,13 @@ deployment:
   #      lines, adjust them as necessary, and remove the curly braces after 'annotations:'.
   #      e.g.  sidecar.istio.io/rewriteAppHTTPProbers: "true"
 
+  # -- Configure horizontal pod autoscaler for deployment
+  autoscaling:
+    enabled: false
+    minReplicas: 1
+    maxReplicas: 100
```
Demonsthere, line 293:
```suggestion
    maxReplicas: 5
```

Just so someone won't say that we recommend 100 replicas 😉

helm/charts/oathkeeper/templates/hpa-controller.yaml
```diff
@@ -0,0 +1,31 @@
+{{- if and (eq .Values.global.ory.oathkeeper.maester.mode "controller") .Values.deployment.autoscaling.enabled }}
```
Demonsthere, line 1: I'm not sure this logic is required, we have a deployment which can be scaled in both cases. And I don't think we should bound hpa to an optional controller. Imho it would be better to have it like this in [hydra](https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/hpa.yaml)

Demonsthere (changes requested):
Hi there! Thanks for the pr, left some minor comments 😉

helm/charts/oathkeeper/templates/hpa-controller.yaml
PeteMac88: I added this because of the if statement in the `deployment-controller.yaml`. I can remove it if you want?

Demonsthere: The if in the `deployment-controller.yaml` controls if we would deploy the `maester` as a sidecar within the main oathkeeper deployment or not, but the end result in still one deployment for oathkeeper. We can safely attach the hpa to the deployment regardless of the maester setup ;)

Demonsthere (approved):
Great! Could you sign the CLA so we can merge it :)?

Issue #467, opened 2022-07-19
Move the CI to GHA


Demonsthere:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently most of the Ory repositories are using GithubActions for CI and release procedures. This repo has not yet been migrated

### Describe your ideal solution

- All pipeline functionalities are migrated to GHA
- Each chart has its own pipeline for testing/upgrading/releasing
- Path filtering is used to only run required actions
- Deprecated elements (like kind) are upgraded or replaced with alternatives (like k3d)

### Workarounds or alternatives

no

### Version

yes

Demonsthere:
Followups:
- [ ] Add secrets to GHA
- [ ] Make a release with the new CI
- [ ] Restore upgrade tests
- [ ] Run CI on changes in the `hack/values.yaml` files
- [ ] Run CI on changes to the CI definitions

PR #472, opened 2022-07-21
ci: migrate to github action (#467)

AlexandreBrg:
#467 

## Checklist

- [ ] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.

## Further comments

I'm using matrix here, as I think it's the best fit for this context & it will easily upgrade when new charts will appear in the repository

- You can find a fully working pipeline with several helm charts update & the push/pr context [here](https://github.com/AlexandreBrg/k8s/actions/runs/2711634085)

- In case of release, the previous pipeline runs and then run the release job, you can find it [here](https://github.com/AlexandreBrg/k8s/actions/runs/2711666502). It creates a commit & push new documentation 

**CircleCI won't pass because there is no `config.yaml` file 🤡**


.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+        run: |
+           updated_charts=$(echo "$FILTER_OUTPUT" | jq -r 'to_entries | map(select((.key | endswith("-helm")) and .value == "true")) | map(.key)')
+           updated_charts=$(echo "$updated_charts" | sed "s/-helm//g")
+
+           if [[ "$updated_charts" == "[]" ]]; then 
+            echo ::set-output name=no_changes::"true"
+           fi
+           echo ::set-output name=matrix::{\"chart\":$(echo $updated_charts)}\"
+           
+           echo "Charts array to run CI on: $updated_charts"
```
AlexandreBrg, line 64: We generate dynamically matrix of jobs for each helm charts updated, this step forge output of `dorny/paths-filter` into something GitHub action matrix can take.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+  
+  check:
+    name: Check Helm Chart '${{ matrix.chart }}'
+    if: ${{ needs.detect-repo-changes.outputs.no-changes != 'true' }}
```
AlexandreBrg, line 68: This line prevents crashes when none helm chart has been updated. It also prevents release to run when no chart updated

.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+
+           if [[ "$updated_charts" == "[]" ]]; then 
+            echo ::set-output name=no_changes::"true"
+           fi
+           echo ::set-output name=matrix::{\"chart\":$(echo $updated_charts)}\"
+           
+           echo "Charts array to run CI on: $updated_charts"
+  
+  check:
+    name: Check Helm Chart '${{ matrix.chart }}'
```
AlexandreBrg, line 67: Either check && upgrade will be a matrix of all helm charts changed in a PR / commit. We can also tell the difference between each job of the matrix thanks to the variable in the name

.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+  #     - name: Checkout
+  #       uses: actions/checkout@v2
+  #     - name: Install k3d 
+  #       run: |
+  #         export PATH=".bin:$PATH"
+  #         curl -s https://raw.githubusercontent.com/k3d-io/k3d/v5.4.4/install.sh | USE_SUDO="false" K3D_INSTALL_DIR=".bin" bash
+  #     - name: Test upgrade for helm chart 
+  #       run: make helm-upgrade
+  #   strategy:
+  #     matrix: ${{ fromJson(needs.detect-repo-changes.outputs.helms-changed) }}
```
AlexandreBrg, line 103: I have put in comment this part, as we cannot make it work right now as there are breaking changes on scripts & helm values for CI. Here what it would do if we run it right now:

```
- Pipeline runs, generate test-upgrade matrix of jobs
- Each test-upgrade jobs tries to fetch script /hacks/values/${HELM_CHART} from the previous version of the chart. <-- At this point it fails because /hacks doesn't exist 
```

This is due to a move of helm CI values and scripts into folder `/hacks`. We will be able to uncomment this job after the next release of charts

.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+        run: make release
+      - name: Push commit for release 
+        run: |
+          git config --global user.email "aeneas@ory.sh"
+          git config --global user.name "aeneasr"
+          git checkout -b make-release HEAD
+          git add -A 
+          git commit -a -m "Release ${RELEASE_VERSION}
+          
+          [skip ci]"
```
AlexandreBrg, line 153: [`[skip-ci]`](https://github.blog/changelog/2021-02-08-github-actions-skip-pull-request-and-push-workflows-with-skip-ci/) prevent having a loop of CI runs :D

.github/workflows/ci.yml
AlexandreBrg: The drawback from it, is at if you don't update any chart you would end up with jobs name unparsed:

image

.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+  #     - name: Test upgrade for helm chart 
+  #       run: make helm-upgrade
+  #   strategy:
+  #     matrix: ${{ fromJson(needs.detect-repo-changes.outputs.helms-changed) }}
+ 
+  test-install:
+    name: Install Helm Chart '${{ matrix.chart }}'
+    if: ${{ needs.detect-repo-changes.outputs.no-changes != 'true' }}
+    needs: [check, detect-repo-changes, dependencies]
+    runs-on: ubuntu-20.04
```
Demonsthere, line 109: Could we specify the image version as a global env, so that we could upgrade/modify it only in 1 place instead?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,155 @@
...
+    # needs: [test-install, test-upgrade]
+    needs: [test-install]
+    runs-on: ubuntu-20.04
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v2
+      - name: Define current tag version
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+
+      - name: Generate documentation
```
Demonsthere, line 136: We'd like docs to be generated not on release, but on post push to the master branch ;)

Demonsthere (changes requested):
Great first version! Leaving some small comments :)

.github/workflows/ci.yml
AlexandreBrg: We can indeed have a dummy entry in the array, but it also means that the job must do nothing when the input chart is this dummy value.

About your alternative solution, it isn't working in `jobs.name`, sadly

PR #484, opened 2022-08-04
fix: use the correct values naming within hydra-maester for consistency 

adamstrawson:
The use of `annotation` within `hydra-maester` is inconsistent with the other Ory Helm Charts, as well as it's own documentation. 

The `hydra-maester` documentation specified `deployment.annotations` already, and the other charts use the same format. 

```
BREAKING CHANGES: This patch changes the behavior of configuration item `annotation`. To keep the existing
behavior please do `deployment.annotation`.
```

## Checklist

- [X] I have added necessary documentation within the code base (if appropriate).

### DIFF
--- a/helm/charts/hydra-maester/templates/deployment.yaml
+++ b/helm/charts/hydra-maester/templates/deployment.yaml
@@ -22,7 +22,7 @@ spec:
         {{- with .Values.deployment.podLabels }}
         {{- toYaml . | nindent 8 }}
         {{- end }}
-      {{- with .Values.annotations }}
+      {{- with .Values.deployment.annotations }}
       annotations:
         {{- toYaml . | nindent 8 }}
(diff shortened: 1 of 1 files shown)


Demonsthere:
Hi there! 
You are correct, the values is not consistent with the rest of the ory ecosystem :/ However your method introduces a breaking change, instead I would suggest the usage of ternary operator:
```yaml
{{- $annotations := ternary .Values.deployment.annotations .Values.deployment.annotation (not (empty .Values.deployment.annotations )) -}}
```

this format would support both and not be a breaking change :) 
Please add in your PR an empty object as the default value for annotations, and some test annotations to the test overrides in the `hack` folder :)

Demonsthere (changes requested):
Please see my comment on the PR :)

Issue #486, opened 2022-08-07
[Kratos] - allow for custom yaml configmap name


Davincible:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Currently, you can only config Kratos through either inline yaml in `values.yaml`, or specify every option through the env variables. Since the config contains info I don't want in VCS, I'm trying to avoid the former, but the latter becomes tedious when you want to specify a lot of custom values.

### Describe your ideal solution

It would be much nicer if just like with the secret, you could provide a custom configmap name that will be used as configuration, so you can in yaml define the whole config, and avoid it going into VCS.

### Workarounds or alternatives

Envs

### Version

latest

Davincible:
Just found out that as a work around I can add an extra `-c file.yaml` and volume mount

Demonsthere:
Hi there! 
If you are concerned about leaking sensitive data within the values/VCS, I would recommend overwriting those values with envs that are read from secrets. this should be more secure then providing a custom cm 😉

Davincible:
Yes but you don't want to write out 20 ENVs. I managed to solve it like this, but the stateful set is missing the `extraArgs` option

```yaml
deployment:
  extraVolumes:
    - name: config-volume
      configMap:
        name: kratos-config-extra
  extraVolumeMounts:
    - name: config-volume
      mountPath: /etc/kratos/
  extraArgs:
    - -c
    - /etc/kratos/config.yaml

statefulset:
  extraVolumes:
    - name: config-volume
      configMap:
        name: kratos-config-extra
  extraVolumeMounts:
    - name: config-volume
      mountPath: /etc/kratos/
  extraArgs:
    - -c
    - /etc/kratos/config.yaml

```

PR #488, opened 2022-08-12
feat(kratos): add statefulset.extraArgs option

Davincible:
Also fix capitalization in readme.md

Fix: #486

## Checklist

- [x] I have added necessary documentation within the code base (if appropriate).

### DIFF
--- a/helm/charts/kratos/README.md
+++ b/helm/charts/kratos/README.md
@@ -100,15 +100,16 @@ A ORY Kratos Helm chart for Kubernetes
 | serviceMonitor.scrapeInterval | string | `"60s"` | Interval at which metrics should be scraped |
 | serviceMonitor.scrapeTimeout | string | `"30s"` | Timeout after which the scrape is ended |
 | serviceMonitor.tlsConfig | object | `{}` | TLS configuration to use when scraping the endpoint |
-| statefulSet.annotations | object | `{}` |  |
-| statefulSet.extraContainers | object | `{}` | If you want to add extra sidecar containers. |
-| statefulSet.extraEnv | list | `[]` |  |
-| statefulSet.extraInitContainers | object | `{}` | If you want to add extra init containers. These are processed before the migration init container. |
-| statefulSet.extraVolumeMounts | list | `[]` |  |
-| statefulSet.extraVolumes | list | `[]` | If you want to mount external volume For example, mount a secret containing Certificate root CA to verify database TLS connection. |
-| statefulSet.labels | object | `{}` |  |
-| statefulSet.nodeSelector | object | `{}` | Node labels for pod assignment. |
-| statefulSet.resources | object | `{}` |  |
+| statefulset.annotations | object | `{}` |  |
+| statefulset.extraContainers | object | `{}` | If you want to add extra sidecar containers. |
+| statefulset.extraEnv | list | `[]` |  |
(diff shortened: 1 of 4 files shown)


helm/charts/kratos/README.md
```diff
@@ -100,15 +100,16 @@ A ORY Kratos Helm chart for Kubernetes
...
-| statefulSet.annotations | object | `{}` |  |
-| statefulSet.extraContainers | object | `{}` | If you want to add extra sidecar containers. |
-| statefulSet.extraEnv | list | `[]` |  |
-| statefulSet.extraInitContainers | object | `{}` | If you want to add extra init containers. These are processed before the migration init container. |
-| statefulSet.extraVolumeMounts | list | `[]` |  |
-| statefulSet.extraVolumes | list | `[]` | If you want to mount external volume For example, mount a secret containing Certificate root CA to verify database TLS connection. |
-| statefulSet.labels | object | `{}` |  |
-| statefulSet.nodeSelector | object | `{}` | Node labels for pod assignment. |
-| statefulSet.resources | object | `{}` |  |
+| statefulset.annotations | object | `{}` |  |
```
Demonsthere, line 103: The docs are auto-generated, no need to modify it :)

helm/charts/kratos/templates/statefulset-mail.yaml
```diff
@@ -54,7 +54,14 @@ spec:
         - name: {{ include "kratos.fullname" . }}-courier
           image: {{ include "kratos.image" . }}
           imagePullPolicy: {{ include "kratos.imagePullPolicy" . }}
-          args: [ "courier", "watch", "--config", "/etc/config/kratos.yaml" ]
+          args:
+            - courier
+            - watch 
+            - --config
+            - /etc/config/kratos.yaml
+            {{- if .Values.statefulset.extraArgs }}
```
Demonsthere, line 62: Please add the new parameter to the values.yaml with a default value of `[]`.

A good idea would be also to add some testing parameters to the values in the `/hack` directory, to ensure that the manifests are properly generated :)

Demonsthere (changes requested):

helm/charts/kratos/README.md
Davincible: I see. I think my changes can just be overwritten with the generated docs right?

Davincible:
I was confused as in the same helm template, both `statefulSet` and `statefulset` were used. I changed all to the former since it aligns more with the rest of the conventions used throughout the template. This is a breaking change, however.

helm/charts/kratos/values.yaml
```diff
@@ -324,6 +324,11 @@ statefulSet:
   #    cpu: 100m
   #  memory: 128Mi
 
+  extraArgs: []
+  # -- Array of extra arguments to be passed down to the StatefulSet. Kubernetes args format is expected
```
Demonsthere, line 328:
```suggestion
  # -- Array of extra arguments to be passed down to the StatefulSet. Kubernetes args format is expected
  extraArgs: []
```

Demonsthere: The helm-docs app we use for docs generation requires the comment be above the variable :)

Demonsthere:
I see, good catch with the `statefulset`, looks like a leftover that was missed when we were refactoring the chart. It is a breaking change, but for the sake of consistency I am in favour of merging it

Demonsthere (approved):

PR #492, opened 2022-08-19
feat: expose Oathkeeper check port

hperl:
This PR adds the option to configure the gRPC Check service for Oathkeeper.

Blocked by https://github.com/ory/oathkeeper/pull/993


Demonsthere:
Hi there! I see that the ci installation fails, maybe we need to bump the default version of oathkeeper?

hperl:
> Hi there! I see that the ci installation fails, maybe we need to bump the default version of oathkeeper?

Yes, the change in Oathkeeper needs to be merged before this, because it complains about unknown config values. I updated the description.

hperl:
Converting to draft, maybe this is not even needed.

PR #504, opened 2022-09-06
feat: allow custom commands and args on migration containers

adamstrawson:
**Note**: Replaces #419, which appears to be inactive/stale. 

Allow the ability to pass through custom commands & args into the migration container. 

For our use case, we run the GCP cloud-sql-proxy sidecar containers, the migrations fail to run as they sometimes start before the proxy has started. This change allows us to add a `sleep` (or a `wait-for-it` script) to the migration command so that it waits for the sql proxy to be available. 

## Checklist

- [ ] I have referenced an issue containing the design document if my change introduces a new feature. **N/A**
- [ ] I have added tests that prove my fix is effective or that my feature works. **N/A** - Tested via `helm template` and `helm install`
- [X] I have added necessary documentation within the code base (if appropriate).

### DIFF
--- a/helm/charts/keto/values.yaml
+++ b/helm/charts/keto/values.yaml
@@ -185,7 +185,9 @@ keto:
     # -- Ability to override the entrypoint of the automigration container
     # (e.g. to source dynamic secrets or export environment dynamic variables)
     customCommand: []
-    #   - sleep 5;
+    # -- Ability to override arguments of the entrypoint. Can be used in-depended of customCommand 
+    # eg:
+    # - sleep 5;
     #   - keto
     customArgs: []
   # -- Direct keto config. Full documentation can be found in https://www.ory.sh/keto/docs/reference/configuration
--- a/helm/charts/keto/values.yaml
+++ b/helm/charts/keto/values.yaml
@@ -182,6 +182,12 @@ keto:
     # When set to initContainer, the migration will be executed when kratos pod is created
     # Defaults to job  
     type: job
+    # -- Ability to override the entrypoint of the automigration container
+    # (e.g. to source dynamic secrets or export environment dynamic variables)
+    customCommand: []
+    #   - sleep 5;
+    #   - keto
+    customArgs: []
   # -- Direct keto config. Full documentation can be found in https://www.ory.sh/keto/docs/reference/configuration
   config:
     serve:
--- a/helm/charts/kratos/values.yaml
+++ b/helm/charts/kratos/values.yaml
@@ -117,7 +117,9 @@ kratos:
     # -- Ability to override the entrypoint of the automigration container
     # (e.g. to source dynamic secrets or export environment dynamic variables)
     customCommand: []
-    #   - sleep 5;
+    # -- Ability to override arguments of the entrypoint. Can be used in-depended of customCommand 
+    # eg:
+    # - sleep 5;
     #   - kratos
     customArgs: []
 
--- a/helm/charts/kratos/values.yaml
+++ b/helm/charts/kratos/values.yaml
@@ -114,6 +114,12 @@ kratos:
     # When set to initContainer, the migration will be executed when kratos pod is created
     # Defaults to job  
     type: job
+    # -- Ability to override the entrypoint of the automigration container
+    # (e.g. to source dynamic secrets or export environment dynamic variables)
+    customCommand: []
+    #   - sleep 5;
+    #   - kratos
+    customArgs: []
 
   # -- You can add multiple identity schemas here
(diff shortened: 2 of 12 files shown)


Demonsthere:
Hi there!
As much as I agree on the core of the issue (how to schedule task that need to be run before the migration), I think this is only a partial solution. 
The application containers are bare bones by design and lack utilities that would allow you for a proper wait-for-x function (like curl), and while adding a simple sleep might be enough, it still leaves us with a possible race condition. 

Moreover, in the case of a sidecar, which runs in parallel to the main application (if I understand your issue correctly), we have a circular dependency as the initContainer needs the sidecar, which won't start before the init finishes. 

Imho the only way to handle a proxy sidecar, would be to disable the automigration job and initContainer, and override the command for the main container to include a wait-for-sidecar then run migrations, then run the main application 😞

adamstrawson:
> Hi there! As much as I agree on the core of the issue (how to schedule task that need to be run before the migration), I think this is only a partial solution. The application containers are bare bones by design and lack utilities that would allow you for a proper wait-for-x function (like curl), and while adding a simple sleep might be enough, it still leaves us with a possible race condition.
> 
> Moreover, in the case of a sidecar, which runs in parallel to the main application (if I understand your issue correctly), we have a circular dependency as the initContainer needs the sidecar, which won't start before the init finishes.
> 
> Imho the only way to handle a proxy sidecar, would be to disable the automigration job and initContainer, and override the command for the main container to include a wait-for-sidecar then run migrations, then run the main application 😞

That's an interesting thought @Demonsthere, and something I didn't consider.  We were actually about to spike moving our own migrations into init containers in the near future, and hadn't crossed out mind about the circular dependency there either. 

I guess if we were to handle migrations with a `job` rather than an `initContainer`, this wouldn't be an issue, as the proxy would run as a sidecar on the job pod as normal. 

Would you accept the PR if it only modified the job resource to pass a custom command/arg, and remove the changes to the `deployment.yaml`? 

For context, this PR was off the back of a previous (stale) one, where I believe others have had similar use-cases not specific to proxies - https://github.com/ory/k8s/pull/419

Demonsthere:
Yeah, a job with the sidecar (and probably a wait-for) would imho work in your case :) Regarding the PR i think we can add the customCommand and args for both app and job, however:
- For consistency, could you add it to all of the applications?
- For testing, add some sample overrides to to the value files in `hack` so that the CI can check if the resulting yamls are valid and can be applied?

adamstrawson:
> For consistency, could you add it to all of the applications?  

All done, looks like it was just Ketos remaining, so all are now consistent

> For testing, add some sample overrides to to the value files in hack so that the CI can check if the resulting yamls are valid and can be applied?

That's done too, I added a mix of `customCommand` and `customArg` across the three applications, giving it a few variations.

Let me know if there's any further additions you'd like

helm/charts/keto/values.yaml
```diff
@@ -182,6 +182,12 @@ keto:
     # When set to initContainer, the migration will be executed when kratos pod is created
     # Defaults to job  
     type: job
+    # -- Ability to override the entrypoint of the automigration container
+    # (e.g. to source dynamic secrets or export environment dynamic variables)
+    customCommand: []
+    #   - sleep 5;
```
Demonsthere, line 188:
```suggestion
    # -- Ability to override arguments of the entrypoint. Can be used in-depended of customCommand 
    # eg:
    # - sleep 5;
```

helm/charts/kratos/values.yaml
```diff
@@ -114,6 +114,12 @@ kratos:
     # When set to initContainer, the migration will be executed when kratos pod is created
     # Defaults to job  
     type: job
+    # -- Ability to override the entrypoint of the automigration container
+    # (e.g. to source dynamic secrets or export environment dynamic variables)
+    customCommand: []
+    #   - sleep 5;
```
Demonsthere, line 120:
```suggestion
    # -- Ability to override arguments of the entrypoint. Can be used in-depended of customCommand 
    # eg:
    # - sleep 5;
```

Demonsthere (approved):

adamstrawson:
Thanks for approving/merging @Demonsthere 🙌  

Just for my own benefit, what's the process around releasing chart versions? Does it tend to happen around app releases, or as and when suits?

Demonsthere:
Chart releases are done by me, and do not require any app releases :) As to when, well, when there is something to release 😅

adamstrawson:
Thanks! I'll just be patient and await for when this PR is released in the next chart version, we can continue to apply migrations manually for the time being

PR #509, opened 2022-09-09
chore: hydra DSN is empty, maester in singleNamespaceMode with Role

roko99:

1. Hydra: to not pass DSN environment variable, when `.Values.hydra.config.dsn` isn't specified. https://github.com/ory/k8s/pull/416
2. Hydra Maester: to not create ClusterRole, when `singleNamespaceMode` is enabled, add relevant permission to Role.
3. Hydra Maester: remove `.Release.Namespace` suffixes for Roles/RoleBindings as they are already per namespace (to unify naming).

## Checklist

## Further comments

DSN change.
When `.Values.hydra.config.dsn` isn't specified - DSN environment variable is passed into the pod with empty value, as it is read from secret, dns key (it is created in any case, with or without value). My suggestion is to not pass DSN environment variable into the pod, when it is empty, and allow user overwrite it with `extraEnv`. If user specifies dsn in `.Values.hydra.config.dsn` it will behave as it does now.

Hydra Maester changes.
- We don't need ClusterRole when `singleNamespaceMode` is enabled, but relevant permissions should be grant to the Role. In our case we are using multiple hydras per environment/namespaces - and if we try to keep unified naming in namespaces, it cause conflicts in ClusterRole names.
- Also I've removed namespace suffixes for Roles/RoleBinding, as they are created per namespace.

### DIFF
--- a/docs/helm/hydra.md
+++ b/docs/helm/hydra.md
@@ -292,6 +292,31 @@ $ hydra token client \
     --client-secret secret
 ```
 
+### Set up DSN variable on runtime
+
+If you use need to construct DSN environment variable on the fly, you can leave
+`hydra.config.dsn` empty and provide custom DSN variable via `extraEnv`, e.g.:
+
+```yaml
+deployment:
+  extraEnv:
+  - name: DB_USER
+    valueFrom:
+      secretKeyRef:
+        name: hydra.postgres-hydra.credentials.postgresql.acid.zalan.do
+        key: username
+  - name: DB_PASSWORD
+    valueFrom:
+      secretKeyRef:
+        name: hydra.postgres-hydra.credentials.postgresql.acid.zalan.do
+        key: password
+  - name: DSN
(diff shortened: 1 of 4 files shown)


Demonsthere (commented):
Hi there,
The change is great in its simplicity :) I am still conflicted regarding passing the empty DSN, as the contract was that we expect it to be provided by the user. This however is still valid as hydra won't start without it, but the way we provide it changes. 

Please add a segment to the [hydra document](https://github.com/ory/k8s/blob/master/docs/helm/hydra.md) with an example :)

roko99:
Hello @Demonsthere
Yes, I agree, that we expect user to provide dsn configuration. And when user doesn't do that application will fail and it is expected. My suggestion is to give people some flexibility with their configuration (basically as my team needs to integrate it with postgres operator :smile:, as well as in related [PR](https://github.com/ory/k8s/pull/416) guys are looking for such opportunity) and not to break existing functionality.

Demonsthere (approved):
LGTM :)

PR #518, opened 2022-09-27
chore: removed / from helm notes because kratos cli cannot handle the /

tikr7:
After installing kratos with helm, the notes say:
```
export KRATOS_PUBLIC_URL=http://127.0.0.1:80/
export KRATOS_ADMIN_URL=http://127.0.0.1:80/
```
The kratos cli cannot handle the last `/` (at least version v0.10.1, v0.10.0 and v0.9.0-alpha.3):
```
kratos ls identities 100 1
2022/09/27 08:57:43 [DEBUG] GET http://127.0.0.1:8081//admin/identities?page=100&per_page=1
```
(See `//`)

Just with:
```
export KRATOS_PUBLIC_URL=http://127.0.0.1:80
export KRATOS_ADMIN_URL=http://127.0.0.1:80
```
it works as expected. So I opened this PR to avoid the same issue for others.
Would be happy about a merge =)


Demonsthere (approved):
Hi there! Good catch with the trailing slashes :) Tanks for the PR!

PR #528, opened 2022-10-14
feat: add support for running Hydra Janitor as a cron job

### LINKED ISSUE #513: Add support for `hydra janitor` in the Hydra Helm chart
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe your problem

Some tables of the database of Ory Hydra keep growing in an unbounded manner, leading to a lot of stale rows that can be cleaned up after some time passed. For this reason, the `hydra janitor` command was born, which can be run as a cron job to purge the data from the database. Unfortunately, the Helm chart does not provide any out-of-the-box way to schedule such job.

### Describe your ideal solution

Provide a `CronJob` resource in the Helm chart that can be enabled and configured via Helm values to execute the `hydra janitor` command. It should be possible to configure how often the cron job runs and which kind of things it should clean up: tokens, requests or grants (any combination of them). It should also be possible to configure the main [options](https://www.ory.sh/docs/hydra/cli/hydra-janitor#options) supported by the command, e.g. the refresh token lifespan or the batch size.

### Workarounds or alternatives

Keep things as-is and leave the implementation of the cron job to the customers

### Version

0.25.3

### Additional Context

I could be able to work on a contribution

ste93cry:

Fixes #513. I added a `CronJob` resource, disabled by default, that can be configured to run `hydra janitor` at a specific interval. I decided to not support all available CLI flags because I don't see the need to customize them. However, I'm open to change my mind if you think that we should make all of them configurable. Below a little table that lists each option of the command and whether it is configurable or not via Helm values:

| CLI flag | Configurable |
| - | - |
| `--access-lifespan` | :x: |
| `--batch-size` | ✅ |
| `--consent-request-lifespan` | :x: |
| `--grants` | ✅ |
| `--keep-if-younger` | :x: |
| `--limit` | ✅ |
| `--refresh-lifespan duration` | :x: |
| `--requests` | ✅|
| `--tokens` | ✅|

I tested all available Helm values related to `janitor` on a development Kubernetes cluster and it seems to work fine.

## Checklist

### DIFF
--- a/helm/charts/hydra/templates/janitor-cron-job.yaml
+++ b/helm/charts/hydra/templates/janitor-cron-job.yaml
@@ -1,0 +1,103 @@
+{{- if .Values.janitor.enabled -}}
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "hydra.fullname" . }}-janitor
+  {{- if .Release.Namespace }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+  labels:
+    {{- include "hydra.janitor.labels" . | nindent 4 }}
+    {{- with .Values.janitor.cronjob.labels }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+  annotations:
+    {{- with .Values.janitor.cronjob.annotations }}
+      {{- toYaml . | nindent 4 }}
+    {{- end }}
+spec:
+  concurrencyPolicy: Forbid
+  schedule: {{ .Values.janitor.cronjob.schedule | quote }}
+  jobTemplate:
+    spec:
+      template:
+        metadata:
+          labels:
+            {{- include "hydra.janitor.labels" . | nindent 12 }}
+            {{- with .Values.janitor.cronjob.labels }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+            {{- with .Values.janitor.cronjob.podMetadata.labels }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          annotations:
+            {{- include "hydra.annotations.checksum" . | nindent 12 -}}
+            {{- with .Values.janitor.cronjob.annotations }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+            {{- with $.Values.janitor.cronjob.podMetadata.annotations }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+        spec:
+          restartPolicy: OnFailure
+          {{- with .Values.imagePullSecrets }}
+          imagePullSecrets:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
+          volumes:
+            - name: {{ include "hydra.name" . }}-config-volume
+              configMap:
+                name: {{ include "hydra.fullname" . }}
+          containers:
+            - name: janitor
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              {{- if .Values.janitor.cronjob.securityContext.enabled }}
+              securityContext:
+                {{- omit .Values.janitor.cronjob.securityContext "enabled" | toYaml | nindent 16 }}
+              {{- end }}
+              command: ["hydra"]
+              args:
+                - janitor
+                {{- if .Values.janitor.cleanupGrants }}
+                - --grants
+                {{- end }}
+                {{- if .Values.janitor.cleanupRequests }}
+                - --requests
+                {{- end }}
+                {{- if .Values.janitor.cleanupTokens }}
+                - --tokens
+                {{- end }}
+                - --batch-size
+                - {{ .Values.janitor.batchSize | quote }}
+                - --limit
+                - {{ .Values.janitor.limit | quote }}
+                - --config
+                - /etc/config/hydra.yaml
+              env:
+              {{- if not (empty ( include "hydra.dsn" . )) }}
+                - name: DSN
+                  valueFrom:
+                    secretKeyRef:
+                      name: {{ include "hydra.secretname" . }}
+                      key: dsn
+              {{- end }}
+              resources:
+                {{- toYaml .Values.janitor.cronjob.resources | nindent 16 }}
+              volumeMounts:
+                - name: {{ include "hydra.name" . }}-config-volume
+                  mountPath: /etc/config
+                  readOnly: true
+          {{- with .Values.janitor.cronjob.nodeSelector }}
+          nodeSelector:
+            {{- toYaml . | nindent 12 }}
+          {{- end }}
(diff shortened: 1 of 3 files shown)


helm/charts/hydra/templates/janitor-cron-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+          containers:
+            - name: janitor
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              {{- if .Values.janitor.cronjob.securityContext.enabled }}
+              securityContext:
+                {{- omit .Values.janitor.cronjob.securityContext "enabled" | toYaml | nindent 16 }}
+              {{- end }}
+              command: ["hydra"]
+              args:
```
Demonsthere, line 60: I'm not sure if this is overcomplicated. Wouldn't it be simpler to set default args in values yaml following the formula:

```yaml
janitor:
  # -- Default args send to the hydra janitor command. You can override this field to input your own arguments
  args:
    - janitor
    - --config
    - /etc/config/hydra.yaml
    - --read-from-env
    - $DSN
  # -- Extra Arguments appended to the hydra janitor command
  extraArgs: []
```

helm/charts/hydra/templates/janitor-cron-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+              configMap:
+                name: {{ include "hydra.fullname" . }}
+          containers:
+            - name: janitor
+              image: "{{ .Values.image.repository }}:{{ .Values.image.tag }}"
+              imagePullPolicy: {{ .Values.image.pullPolicy }}
+              {{- if .Values.janitor.cronjob.securityContext.enabled }}
+              securityContext:
+                {{- omit .Values.janitor.cronjob.securityContext "enabled" | toYaml | nindent 16 }}
+              {{- end }}
```
Demonsthere, line 58: I would replace the block with a simple `{{- with .Values.janitor.cronjob.securityContext }}`, which won't render if the object is empty. Unless we want to specify some defaultContext, in which case we need to update the values.yaml 😉

helm/charts/hydra/templates/janitor-cron-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+apiVersion: batch/v1
+kind: CronJob
+metadata:
+  name: {{ include "hydra.fullname" . }}-janitor
+  {{- if .Release.Namespace }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
+  labels:
+    {{- include "hydra.janitor.labels" . | nindent 4 }}
+    {{- with .Values.janitor.cronjob.labels }}
```
Demonsthere, line 11: Just as a general note, we tend to separate application logic from deployment logic, that is why `deployment` and `jobs` have dedicated top-level field. I would prefer to have 

```
---
# values.yaml
janitor:
  applicationLogicFields

cronjob:
  janitor:
    k8sCronSettings
```

Demonsthere (commented):
Left a few comments, but overall good job :)

Demonsthere:
Also, we might want to set the values in `hack/values` to enable the cronjob so it will be rendered by the CI and checked against manifest validation :)

helm/charts/hydra/templates/janitor-cron-job.yaml
ste93cry: >Unless we want to specify some defaultContext, in which case we need to update the values.yaml

Do you think we should? I see that the deployment has a default value configured, but I don't know what the best practice is for Helm charts out in the world.

helm/charts/hydra/templates/janitor-cron-job.yaml
ste93cry: I wanted to structure the chart values and avoid leaving this freedom to the user in terms of command line customization. However, if you think it is too rigid a thought, I have nothing negative about changing my mind.

helm/charts/hydra/templates/janitor-cron-job.yaml
Demonsthere: I'd say, since we control the binary and how we deploy, we might want to include default Context. We know the required permissions :) IMHO, we should set them as the same as we have on the deployment

helm/charts/hydra/templates/janitor-cron-job.yaml
Demonsthere: I am not against this structure, but think we should leave the possibility for customisation :) How about leaving the structure, which most probably will be enough for most, but for power users allow a `customArgs` which would override the whole array. What do you think?

helm/charts/hydra/templates/janitor-cron-job.yaml
```diff
@@ -0,0 +1,103 @@
...
+          annotations:
+            {{- include "hydra.annotations.checksum" . | nindent 12 -}}
+            {{- with .Values.cronjob.janitor.annotations }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+            {{- with $.Values.cronjob.janitor.podMetadata.annotations }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+        spec:
+          restartPolicy: OnFailure
```
Demonsthere, line 42: If we want to leave this policy as a default, i'd prefer we also set a default backoffLimit (5 or 10)

helm/charts/hydra/values.yaml
```diff
@@ -394,6 +394,71 @@ watcher:
...
+
+  ## -- Configure if the access and refresh tokens must be cleaned up
+  cleanupTokens: false
+
+  ## -- Configure how many records are deleted with each iteration
+  batchSize: 100
+
+  ## -- Configure how many records are retrieved from database for deletion
+  limit: 10000
+
```
Demonsthere, line 416:
```suggestion
# -- Configuration for cronjobs
```

helm/charts/hydra/templates/janitor-cron-job.yaml
ste93cry: Considering that `customArgs` is already provided for the `Job` that runs the migrations, it makes sense 😄

Demonsthere (approved):
Good job :)

Issue #529, opened 2022-10-20
Race condition on job ServiceAccounts can cause job to fail to run 


adamstrawson:
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Network](https://www.ory.sh/) project.
- [X] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [X] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

We've found that on new Kubernetes cluster, there can be a race condition where the `Job` resource is created before the ServiceAccount exists, causing the job to fail to run. This is especially an issue when using ServiceAccount annotations for IAM permissions.

### Reproducing the bug

1. Create a new cluster
2. Enable job based migrations
3. Install chart 

### Version

v0.25.6

PR #530, opened 2022-10-20
fix: helm hooks for job Service Accounts

adamstrawson:

We've found that on new Kubernetes cluster, there can be a race condition where the Job resource is created before the ServiceAccount exists, causing the job to fail to run. This is especially an issue when using ServiceAccount annotations for IAM permissions.

Issue: #529

By adding helm hooks, we can ensure that the ServiceAccount is applied as early as possible. 

## Checklist

### DIFF
--- a/helm/charts/hydra/templates/job-rbac.yaml
+++ b/helm/charts/hydra/templates/job-rbac.yaml
@@ -5,9 +5,12 @@ kind: ServiceAccount
 metadata:
   name: {{ include "hydra.job.serviceAccountName" . }}
   labels:
-    {{- include "hydra.labels" . | nindent 4 }}
-  {{- with .Values.job.serviceAccount.annotations }}
+    {{- include "hydra.labels" . | nindent 4 }} 
   annotations:
+    helm.sh/hook-weight: "0"
+    helm.sh/hook: "pre-install, pre-upgrade"
+    helm.sh/hook-delete-policy: "before-hook-creation"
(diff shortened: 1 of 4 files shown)


Demonsthere (commented):
Hi there!
I'm not convinced with the content of this PR 😞 
As much as I understand your issue, please not that even in our CI we always run in a from scratch scenario, yet we did not encounter the race condition. 
We also expose the annotations via `.Values.deployment.serviceAccount.annotations` which you can use to add them by yourself :)

adamstrawson:
> Hi there! I'm not convinced with the content of this PR 😞 As much as I understand your issue, please not that even in our CI we always run in a from scratch scenario, yet we did not encounter the race condition. We also expose the annotations via `.Values.deployment.serviceAccount.annotations` which you can use to add them by yourself :)

Hey @Demonsthere  Not a problem, we thought it might be of use to others too, but we'll continue to use the annotations within `.Values.deployment.serviceAccount.annotations` instead.  

I do think this is probably just unique to those using IAM annotations specifically, where permissions are required to run certain actions.

Demonsthere:
> I do think this is probably just unique to those using IAM annotations specifically, where permissions are required to run certain actions.

Yeah I agree on that, but i am just not 100% sure we should have that as default values. Some users just use `helm template | kubectl apply` or other tooling that does not support hooks directly. That is why i restrain from using them unless we don't have any alternatives. 

In this case I would suggesting changing the PR to update the docs and put the values there, which might be helpful to others :)

PR #549, opened 2022-11-22
fix: notes for example-idp in ingress mode

SozinM:
## Description

There is a bug in the NOTES.txt that corrupts shown address. 
To reproduce this bug you need to install/upgrade release with values
```
ingress:
  enabled: true
```

Previously notes used the whole path structure instead of the path itself.  It was leading to an incorrect path shown after helm install/upgrade. 
The path before changes: 
```
1. Get the application URL by running these commands:
  https://example-idp.int.chainstack.commap[path:/ pathType:ImplementationSpecific]
```
The path after changes:
```
1. Get the application URL by running these commands:
  https://example-idp.int.chainstack.com/
```

## Checklist

### DIFF
--- a/helm/charts/example-idp/templates/NOTES.txt
+++ b/helm/charts/example-idp/templates/NOTES.txt
@@ -2,7 +2,7 @@
 {{- if .Values.ingress.enabled }}
 {{- range $host := .Values.ingress.hosts }}
   {{- range .paths }}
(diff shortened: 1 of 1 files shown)


Demonsthere (approved):

Demonsthere:
Good catch :) probably a leftover from the older version of the ingress object

PR #560, opened 2023-01-27
fix: [hydra helm chart] add missing securityContext in initcontainer

monotek:
Signed-off-by: André Bauer <andre.bauer@staffbase.com>

* added securityContext & podSecurityContext to all containers
  * used  {{ with }} instead of {{ if }} for all occurences for consistency reasons
  * migration inti container uses same securityContext & podSecurityContext as hydra
* added resource requests and limit to automigrate initcontainer

## Checklist

### DIFF
--- a/helm/charts/hydra/Chart.yaml
+++ b/helm/charts/hydra/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "v2.0.3"
 description: A Helm chart for deploying ORY Hydra in Kubernetes
 name: hydra
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-hydra.svg
-version: 0.27.1
+version: 0.27.0
 keywords:
   - oauth2
   - openid-connect
--- a/helm/charts/hydra/Chart.yaml
+++ b/helm/charts/hydra/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "v2.0.3"
 description: A Helm chart for deploying ORY Hydra in Kubernetes
 name: hydra
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-hydra.svg
-version: 0.27.0
+version: 0.27.1
 keywords:
   - oauth2
   - openid-connect
--- a/helm/charts/hydra/templates/deployment.yaml
+++ b/helm/charts/hydra/templates/deployment.yaml
@@ -191,6 +191,10 @@ spec:
             {{- with $migrationExtraEnv }}
               {{- toYaml . | nindent 12 }}
             {{- end }}
+          {{- if .Values.deployment.securityContext }}
+          securityContext:
+            {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+          {{- end }}
       {{- end }}
       {{- if .Values.deployment.extraInitContainers }}
         {{- tpl .Values.deployment.extraInitContainers . | nindent 8 }}
--- a/helm/charts/hydra/templates/deployment.yaml
+++ b/helm/charts/hydra/templates/deployment.yaml
@@ -195,9 +195,9 @@ spec:
           resources:
             {{- .Values.hydra.automigration.resources | nindent 12 }}
           {{- end }}
-          {{- if .Values.deployment.securityContext }}
+          {{- with .Values.deployment.initContainerSecurityContext }}
           securityContext:
-            {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+            {{- toYaml . | nindent 12 }}
           {{- end }}
       {{- end }}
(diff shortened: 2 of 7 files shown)


helm/charts/hydra/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "v2.0.3"
 description: A Helm chart for deploying ORY Hydra in Kubernetes
 name: hydra
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-hydra.svg
-version: 0.27.0
+version: 0.27.1
```
Demonsthere, line 6: This is auto-generated, please do not modify :)

helm/charts/hydra/templates/deployment.yaml
```diff
@@ -191,6 +191,10 @@ spec:
             {{- with $migrationExtraEnv }}
               {{- toYaml . | nindent 12 }}
             {{- end }}
+          {{- if .Values.deployment.securityContext }}
```
Demonsthere, line 194: I think this would require a small refactor. Right now we are using `.Values.deployment.securityContext` to set the context for the main application, the init container may have/require different permissions than the main application. I think we would need to split it in the following way:

```yaml
.Values.deployment.securityContext #Pod level context, applies to all containers in the generated pod
.Values.hydra.securityContext #Container level context, applies to the main hydra container
.Values.job.securityContext #Container level context, applies to the init container
```

Demonsthere (commented):
Hello there, and thanks for the PR :) 
Please read my comments, as I think this requires a bit more of a generic solution that could be then applied to all charts in the repo :)

helm/charts/hydra/templates/deployment.yaml
monotek: The migration job alread uses the deployments securityContext: https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/job-migration.yaml#L78

The cronjob has defined its own securityContext: https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/janitor-cron-job.yaml#L55

If we want to add a podSecurityContext, we likly have to add that for janitor cronjob and migration job too?

So i guess, to not to change current values an introducing a braking change, it would have to look like:

```
.Values.deployment.initContainerSecurityContext
.Values.deployment.podSecurityContext
.Values.cronjob.janitor,podSecurityContext
```

What do you think?

Demonsthere: I see your point, and I agree we don't want to introduce breaking changes if we **really** don't need to :) 
In your proposal, would we leave `.Values.deployment.securityContext`? And if so, would that be not pod level (the name would suggest that is applied to the whole deployment) 🤔

monotek: I'v did a first shot on that (also added resources to the initcontainer).

I stumpled upon: https://github.com/ory/k8s/blob/master/helm/charts/hydra/templates/deployment-watcher.yaml#L46

It uses `.Values.securityContext`, which does not exist in the values.yaml, so i've considered it a bug.
I've changed it to `.Values.watcher.securityContext` and added also `.Values.watcher.podSecurityContext`.

Please tell me if this is wanted, as it's changing for users who actually using it this way?


Yes, i would keep `.Values.deployment.securityContext`. I've names all other vars `podSecurityContext` for that reason.

Demonsthere: Yes, `.Values.securityContext` was the first iteration when the chats only deployed the single deployment, this must have been an oversight and orphaned :( I think your changes make sense here, and they seem consistent with the current approach :)

monotek: Let me know if there are any more changes needed and of commits should be squashed before merge?

Demonsthere (commented):
Great job! :) 
Just one more request from me: please add some example values to the `hacks/values/hydra.yaml` which is used by the CI. This will allow us to test the generated manifests in the CI, and check if they are compliant with the k8s api (helpful to detect invalid indents) :)

monotek:
> Great job! :) Just one more request from me: please add some example values to the `hacks/values/hydra.yaml` which is used by the CI. This will allow us to test the generated manifests in the CI, and check if they are compliant with the k8s api (helpful to detect invalid indents) :)

Done :)

Demonsthere (approved):
LGTM! 🙏

vinckr:
Hello @monotek  
Congrats on merging your first PR in Ory 🎉 !
Your contribution will soon be helping secure millions of identities around the globe 🌏. 
As a small token of appreciation we send our first time contributors a gift package to welcome them to the community. 
[Please drop me an email and I will forward you the form to claim your Ory swag!](mailto:vincent@ory.sh)

PR #578, opened 2023-03-15
feat: add loadBalancerIP

adamwalach:
## Checklist

### DIFF
--- a/helm/charts/hydra/templates/service-admin.yaml
+++ b/helm/charts/hydra/templates/service-admin.yaml
@@ -19,6 +19,11 @@ metadata:
     {{- end }}
 spec:
   type: {{ .Values.service.admin.type }}
+  {{- if eq .Values.service.admin.type "LoadBalancer" }}
+  {{- with .Values.service.admin.loadBalancerIP }}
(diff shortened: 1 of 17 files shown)


helm/charts/hydra/templates/service-admin.yaml
```diff
@@ -19,6 +19,11 @@ metadata:
     {{- end }}
 spec:
   type: {{ .Values.service.admin.type }}
+  {{- if eq .Values.service.admin.type "LoadBalancer" }}
+  {{- with .Values.service.admin.loadBalancerIP }}
```
piotrmsc, line 23: will it work for dynamic IP addresses if one want to uses svc of type LB but without reserving staticly IP?

Demonsthere: Should be fine, if no ip is defined it will use the empty string value, which will be then overwritten by the value recovered from status

adamwalach: yes, if the value of loadBalancerIP is empty it doesn't render the line

Demonsthere (approved):

PR #579, opened 2023-03-20
fix: allow deployment strategy override in oathkeeper

razielgn:
Allow customizing deployment strategy.

This is not a breaking change, as the default fallback on values.yaml is the default K8s behaviour.

## Checklist


helm/charts/oathkeeper/values.yaml
```diff
@@ -193,6 +193,12 @@ secret:
   hashSumEnabled: true
 
 deployment:
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
```
Demonsthere, line 198: I think a better approach here would be to have an empty object as a default value, to not enforce/provide concrete values. 
```suggestion
    rollingUpdate: {}
```

Demonsthere: On the other hand, please also update the values in the `hack/values` directory with example values, and so that the CI will use and render them :)

Demonsthere (changes requested):

helm/charts/oathkeeper/values.yaml
razielgn: Sorry for the delay. Right, done! Also see the other PR.

Demonsthere (approved):

PR #601, opened 2023-04-28
feat: parameterise busybox image repo/tag

fenech:
Allow the charts to be used in environments with constraints on connectivity to container image repositories.

A hard-coded name `busybox` means that the Chart tests rely on the Kubernetes cluster being able to pull from Docker Hub. We already allow the component image to be customised, so this follows the same pattern for the busybox image.

## Checklist

### DIFF
--- a/helm/charts/hydra/values.yaml
+++ b/helm/charts/hydra/values.yaml
@@ -566,3 +566,8 @@ configmap:
 test:
   # -- Provide additional labels to the test pod
   labels: {}
+
+# -- use a busybox image from another repository
+busybox:
+  repository: busybox
+  tag: 1
--- a/helm/charts/hydra/values.yaml
+++ b/helm/charts/hydra/values.yaml
@@ -566,8 +566,7 @@ configmap:
 test:
   # -- Provide additional labels to the test pod
   labels: {}
-
-# -- use a busybox image from another repository
-busybox:
-  repository: busybox
-  tag: 1
+  # -- use a busybox image from another repository
+  busybox:
+    repository: busybox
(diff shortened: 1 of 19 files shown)


helm/charts/hydra/values.yaml
```diff
@@ -566,3 +566,8 @@ configmap:
 test:
   # -- Provide additional labels to the test pod
   labels: {}
+
+# -- use a busybox image from another repository
+busybox:
+  repository: busybox
+  tag: 1
```
Demonsthere, line 573:
```suggestion
test:
  # -- Provide additional labels to the test pod
  labels: {}
  # -- use a busybox image from another repository
  image:
    repository: busybox
    tag: 1
```

Demonsthere: Hi there, i think this format would work better, let's keep the test image option in the test object 😉

Demonsthere (changes requested):
Please also set some test override for the image in the `hacks` directory :)

helm/charts/hydra/values.yaml
fenech: Good idea, the only exception to the rule would be in `helm/charts/oathkeeper/templates/deployment-sidecar.yaml`, where the busybox image is also used. What do you think?

Demonsthere: Ah, there we use the image as an initcontainer, if you can please make it configurable in the `image` section like 
```yaml
image:
  initContainer:
    repository: foo
    tag: bar
```

Demonsthere (approved):

PR #608, opened 2023-05-18
fix: run `extraInitContainers` before migration

cschmatzler:
In [`values.yaml`](https://github.com/ory/k8s/blob/1d4a1787b1bb5faa0893f281a537b0ee358d5576/helm/charts/kratos/values.yaml#LL245C3-L246C26), the `deployment.extraInitContainers` value is documented as follows:

> These are processed before the migration init container.

In the template itself, they were added after the migration container, though. Since Kubernetes runs init containers in the order they were defined, and each one of them has to be successful in order for the next to be started, this can block setups where there's an extra init container that sets up the database.

See also the relevant [Kubernetes documentation](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#differences-from-regular-containers):

> If you specify multiple init containers for a Pod, kubelet runs each init container sequentially. Each init container must succeed before the next can run. When all of the init containers have run to completion, kubelet initializes the application containers for the Pod and runs them as usual.

Init containers were added in an order that contradicts the documentation in `values.yaml`.

## Checklist


helm/charts/kratos/templates/deployment-kratos.yaml
```diff
@@ -52,6 +52,9 @@ spec:
         {{- toYaml . | nindent 8 }}
     {{- end }}
       initContainers:
+      {{- if .Values.deployment.extraInitContainers  }}
+{{ tpl  .Values.deployment.extraInitContainers . | indent 8 }}
+      {{- end }}
```
Demonsthere, line 57:
```suggestion
      {{- if .Values.deployment.extraInitContainers  }}
        {{- tpl  .Values.deployment.extraInitContainers . | nindent 8 }}
      {{- end }}
```

Demonsthere: Please use `nindent ` here for more clarity :)

Demonsthere (changes requested):
Hi there!
Good catch, we want to allow setting init containers before the migration, as one of the most popular use case is hydrating credentials to the DB before the migration runs. 

Could you also verify if the same happens for other charts?

cschmatzler:
Hey!
Thanks for the quick review. I have applied your suggestion and also applied the same change to the Keto and Hydra charts.

Let me know if I missed anything, since I'm not too familiar with all the components yet.

Cheers!

Demonsthere (approved):
Thanks for the update! :) 
Only keto, kratos and hydra use a db so we got them all. I think we can merge this as is right now, i might need to refactor it a bit later, as some use cases may require to run after migration, so we might need to split it to `extraInitContainersBeforeMigration` and `extraInitContainersAfterMigration`

PR #610, opened 2023-05-21
docs: update hydra.md

mesat:
This PR fixes a bug in the documentation of the Helm chart for Hydra. The bug occurs when generating a random secret using the command:

`LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | base64 | head -c32`

This command may fail with the following error:

```
error writing output file

/usr/bin/base64:I/O error on input
```

This error is because the `base64` command expects a multiple of 4 bytes as input, but the `tr` command may produce a different number of bytes. To fix this, the order of the commands should be changed so that the `head` command is applied before the `base64` command, like this:

`LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | head -c 32 | base64`

This ensures that the input to the `base64` command is always 32 bytes, a multiple of 4. This command works as expected and generates a random secret.

This PR updates the documentation to use the correct command in 5 places where it occurs. This PR does not contain any code changes, only documentation updates.

## Checklist

### DIFF
--- a/docs/helm/hydra.md
+++ b/docs/helm/hydra.md
@@ -34,7 +34,7 @@ cryptographically strong secret, a Login and Consent provider located at
 
 ```bash
 $ helm install \
-    --set 'hydra.config.secrets.system={$(LC_ALL=C tr -dc 'A-Za-z0-9' < /dev/urandom | base64 | head -c 32)}' \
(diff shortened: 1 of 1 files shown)


Demonsthere (approved):
Hello there!
Thanks for noticing and sending the PR :)

PR #614, opened 2023-05-31
feat: parameterise command and args

xiadaniGarcia:
Adding customCommand and customArgs to the main containers for both Hydra and Keto, these changes will allow us to export dynamic variables 
Thanks
## Checklist

- [X] I have added tests that prove my fix is effective or that my feature works.


xiadaniGarcia:
I have already signed the CLA but it is not being reflected

helm/charts/hydra/templates/deployment.yaml
```diff
@@ -64,14 +64,14 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
+          {{- if .Values.hydra.customCommand }}
```
Demonsthere, line 67: Hi there! 
I'd prefer a slightly different format:
`Values.hydra.command` with the default value of `hydra` and `Values.hydra.args` with the default value of `[]`. This will be more verbose and allows us to drop the if/else clauses :)

helm/charts/keto/templates/deployment.yaml
```diff
@@ -111,11 +111,19 @@ spec:
           {{- end }}
           image: "{{ .Values.image.repository }}:{{ .Values.image.tag | default .Chart.AppVersion }}"
           imagePullPolicy: {{ .Values.image.pullPolicy }}
-          command: [ "keto" ]
+          {{- if .Values.keto.customCommand }}
```
Demonsthere, line 114: Same as with hydra :)

Demonsthere (changes requested):

Demonsthere:
The CLA has some issue :( Could you try signing it again?

helm/charts/hydra/templates/deployment.yaml
xiadaniGarcia: Sure, I agree with having `Values.hydra.command` to avoid the if/else clauses! Do you think is a good  idea to set `Values.hydra.args` default to `[]` when the previous version was like this: 
```
          args:
            - serve
            - all
            {{- if .Values.hydra.dev }}
            - "--dev"
            {{- end }}
            - --config
            - /etc/config/hydra.yaml
```

Demonsthere: Ah true! Forgot about that bit 😅.
For args i think it makes sense to leave customArgs as you proposed as that will override the whole structure. In that case we would need to make it like 

```yaml
{{- if Values.hydra.customArgs }}
args:
...
{{- else }}
          args:
            - serve
            - all
            {{- if .Values.hydra.dev }}
            - "--dev"
            {{- end }}
            - --config
            - /etc/config/hydra.yaml
{{- end }}

Demonsthere:
About the CLA, the problem is that your commits do not have an email assigned to it, that why the bot cannot recognise it :( Could you please add your email to the commits and amend those you've pushed?

helm/charts/hydra/templates/deployment.yaml
xiadaniGarcia: I made the chages :)

helm/charts/keto/templates/deployment.yaml
xiadaniGarcia: Done

Demonsthere (approved):
LGTM :)

PR #619, opened 2023-06-19
fix: make migration extraEnv work through job.extraEnv

Joibel:
This is only a bugfix.

Where job.extraEnv is specified it should be used in preference to deployment.extraEnv, which was not happening.

If deployment.extraEnv had been specified before it would have always been used. Previously the chart wouldn't render at all with job.extraEnv specified, but deployment.extraEnv not specified.

Tested by specifying job.extraEnv and seeing in the output of `helm template .`

## Checklist


Demonsthere (approved):
This looks like a common case of copy&paste error, good work catching it :)

PR #638, opened 2023-09-07
feat: added priorityClassName to all charts

WatcherWhale:
At the moment, only the hydra and hydra-maester charts allow configuring the priorityClassName. This pull request introduces this configuration option to the other charts as well.

## Checklist

### DIFF
--- a/helm/charts/keto/templates/deployment.yaml
+++ b/helm/charts/keto/templates/deployment.yaml
@@ -202,6 +202,9 @@ spec:
       {{- with $extraVolumes }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- with .Values.priorityClassName }}
+      priorityClassName: {{ . }}
+      {{- end }}
       {{- with $nodeSelector }}
       nodeSelector:
         {{- toYaml . | nindent 8 }}
--- a/helm/charts/kratos/values.yaml
+++ b/helm/charts/kratos/values.yaml
@@ -288,6 +288,10 @@ deployment:
   #    cpu: 100m
   #  memory: 128Mi
 
+  # -- Pod priority
+  # https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+  priorityClassName: ""
+
   # -- Node labels for pod assignment.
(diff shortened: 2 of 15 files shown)


helm/charts/kratos/values.yaml
```diff
@@ -425,6 +429,10 @@ statefulSet:
   #      lines, adjust them as necessary, and remove the curly braces after 'labels:'.
   #      e.g.  type: app
 
+  # -- Pod priority
+  # https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+  # priorityClassName: ""
```
Demonsthere, line 434:
```suggestion
  priorityClassName: ""
```

helm/charts/keto/templates/deployment.yaml
```diff
@@ -202,6 +202,9 @@ spec:
       {{- with $extraVolumes }}
         {{- toYaml . | nindent 8 }}
       {{- end }}
+      {{- if .Values.priorityClassName }}
```
Demonsthere, line 205: For internal consistency please use the `with` method instead of `if` :)

helm/charts/kratos/values.yaml
Demonsthere: empty string is evaluated to false in both if and with clauses, so we can safely use it as a default value :)

helm/charts/oathkeeper/values.yaml
```diff
@@ -29,6 +29,10 @@ sidecar:
     tag: v0.1.2
   envs: {}
 
+# -- Pod priority
+# https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/
+priorityClassName: ""
```
Demonsthere, line 34: also please update the values.yaml files in the `hack/values` directory to set those fields. This allows the CI to verify if they generate valid manifest files

Demonsthere (changes requested):
Hello there!
Thanks for the PR, just adding some housekeeping, and then we can merge :)

Demonsthere (approved):
LGTM!

PR #656, opened 2023-11-29
fix: Fix PDB indentation issue

achedeuzot:
I've fixed the PodDisruptionBudget yaml inclusion which would break files by setting fields at the root of the file.

On our  clusters, the helm charts create the following PDBs:

```
metadata:
  name: hydra
  namespace: hydra
minAvailable: 1 # This field is set here and is invalid.
spec:
  maxUnavailable: 25%
  selector:
    matchLabels:
      app.kubernetes.io/instance: hydra
      app.kubernetes.io/name: hydra
status:
```

## Checklist

## Further comments
None


Demonsthere (commented):
Hey there!
In order to properly fix this, please update the values in `hacks/values`, to enable PDB. This will force our CI to render the manifests and validate them. As PDB are optional objects we probably dont enable them by default and that is why this was omitted.

achedeuzot:
I've added the test values. They seem to show exactly one of the errors we had when trying to fix this: the default value is the following:
```
pdb:
  enabled: false
  spec:
    minAvailable: 1
```

When we specify the following:
```
pdb:
  enabled: true
  spec:
    maxUnavailable: 25%
```
The configuration that is computed it wrong, it merges the `spec` leading to 
```
pdb:
  enabled: true
  spec:
    minAvailable: 1
    maxUnavailable: 25%
```
which is not valid. Before my change, it used to work because the generated file would look like this:
```
pdb:
  enabled: true
  spec:
    maxUnavailable: 25%
minAvailable: 1 # Notice the ignored key here
```

How do you suggest I fix this ? I'm not sure what's the best approach ? Thanks !

achedeuzot:
One of the options is to use this instead of the `toYaml`:
```
  {{- if .Values.pdb.spec.maxUnavailable -}}
  maxUnavailable: {{ .Values.pdb.spec.maxUnavailable }}
  {{- else -}}
  minAvailable: {{ .Values.pdb.spec.minAvailable }}
  {{- end -}}
  ```

**Comment by Demonsthere** [comment]
Since the PDB object has only 3 fields (selector, max and min) i think we can remove the `toYaml` and instead expose the fields implicit 

```yaml
spec:
  minAvailable: {{ .Values.pdb.spec.minAvailable }}
  maxUnavailable: {{ .Values.pdb.spec.maxUnavailable }}
```

achedeuzot:
You'll see I put a if/else because you can't specify both.
All checks should be green :) 
There's 1 failing check which is unrelated to this branch.

achedeuzot:
It should be ready to merge @Demonsthere ;)

helm/charts/hydra-maester/templates/pdb.yaml
```diff
@@ -10,5 +10,9 @@ spec:
       control-plane: controller-manager
       app.kubernetes.io/name: {{ include "hydra-maester.fullname" . }}
       app.kubernetes.io/instance: {{ .Release.Name }}
-  {{ toYaml .Values.pdb.spec }}
+  {{- if .Values.pdb.spec.maxUnavailable }}
+  maxUnavailable: {{ .Values.pdb.spec.maxUnavailable }}
+  {{- else }}
+  minAvailable: {{ .Values.pdb.spec.minAvailable | default "1" }}
```
Demonsthere, line 16: i dont like this hidden default value, maybe the following form would be more explicit?

```yaml
# values.yaml
pdb:
  enabled: true
  spec:
    maxUnavailable: ""
    minAvailable: ""
---
# template.yaml
{{- with .Values.pdb.spec.maxUnavailable }}
maxUnavailable: {{ . }}
{{- end }}
{{- with .Values.pdb.spec.minAvailable }}
minAvailable: {{ . }}
{{- end }}
```

I know the spec allows for only one field to be set, but `with` won't render empty fields, and imho it is the users responsibility to give sane data to the system

Demonsthere (changes requested):
Please run `make format` to make the CI happy :) 
I also would prefer a more explicit approach to the values

achedeuzot:
Hi @Demonsthere :)

I've replaced it with `with` instead of `if/else`, nice suggestion :D 

The `make format` doesn't seem to fix the docs file issue (which I haven't touched so it should be valid if it's valid on master... I don't get why the CI fails on a file that wasn't changed).

Demonsthere (approved):
Fixed the formatting, the error was on master branch 😓 
Thanks for the PR! :)

PR #669, opened 2024-03-27
fix: ensure oathkeeper controller deployment correctly account for optionally created resources.

cbrendanprice:
This fix ensures that the oathkeeper controller deployment template correctly accounts for optional creation of configmap and secrets based on the provided values:
- `oathkeeper.secret.enabled`
- `secret.enabled`

n/a (couldn't find an open/closed issue for the controller failing to initialize due to configmap missing for rules)

## Checklist
- [X] I have referenced an issue containing the design document if my change introduces a new feature.

## Further comments
this pr assumed that it should be possible to deploy the maester as a controller whilst also allowing it to manage the access rules. if this is incorrect, im happy to adjust the logic to ensure that the configmap for access rules get created unles *both* `managedAccessRules` is `false` and the maester mode is `sidecar`.

also please let me know if i need to increment the chart's semver patch for fixes; wasn't clear in the checklist above.

### DIFF
--- a/helm/charts/oathkeeper/templates/deployment-controller.yaml
+++ b/helm/charts/oathkeeper/templates/deployment-controller.yaml
@@ -62,14 +62,38 @@ spec:
             name: {{ include "oathkeeper.fullname" . }}-config
             {{- end }}
         - name: {{ include "oathkeeper.name" . }}-rules-volume
+          {{- if .Values.oathkeeper.managedAccessRules }}
           configMap:
             name: {{ include "oathkeeper.fullname" . }}-rules
+          {{- else }}
+          emptyDir: {}
+          {{- end }}
+        {{- if .Values.secret.enabled }}
         - name: {{ include "oathkeeper.name" . }}-secrets-volume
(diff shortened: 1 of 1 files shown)


Demonsthere (commented):
Hi there :) I think this issue was mentioned in the past. I have updated the PR to satisfy kubescape, besides that i think we can merge as is. 
The idea was that oathkeeper as a application requires some rules to be present for it to start properly. In an k8s environment we dont want the deployment to be restarted until some CR is created, therefore we created the empty file to satisfy the startup conditions.

Demonsthere (approved):

edeustua:
Hi there! I recently upgraded the chart and found that it is no longer possible to load a configmap with access rules due to this PR. Was this the intended behavior? If `managedAccessRules` is turned off there is no way to load a custom templated configmap into the deployment. Maybe I'm missing something.

edeustua:
After some thought  I think the write solution might be to set this to only work if maester is enabled.

PR #675, opened 2024-04-25
fix: `hydra-maester.adminService.port` empty value is passed into child chart, causes error.

alangilmor:
Helm chart 0.41.0 requires commenting out the `hydra-maester.adminService.port` value. 

Test: 

`helm template ory/hydra` (port empty in maester Deployment args)

`helm template ory/hydra --set hydra-maester.adminService.port=null`  (child chart's default respected when "unset" in parent chart)


alangilmor (commented):
Hi apologies if the PR is not conform to standards.  Just trying to use the chart and had a failure out-of-the-box.  Adjust as needed.

Demonsthere (approved):
Hi there! Sorry for the long delay 😞 
I have checked the code, and I think we can merge this as is, the issue was the the value in hydra chart in fact did override the value in the subchart, which caused the error

PR #684, opened 2024-05-17
fix: mount secrets to initContainer as well

meysam81:
## Checklist

### DIFF
--- a/helm/charts/kratos/templates/job-migration.yaml
+++ b/helm/charts/kratos/templates/job-migration.yaml
@@ -71,7 +71,7 @@ spec:
             {{- toYaml $migrationExtraEnv | nindent 10 }}
           {{- end }}
         {{- if .Values.deployment.environmentSecretsName }}
-          envFrom:
+        envFrom:
           - secretRef:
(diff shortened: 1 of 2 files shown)


Demonsthere:
Hi there!
I have no real issue with the changes, just one question: do you think that we should use the same envFrom clause in the case of initJob? The initcontainer is an option,. but not the default, and it would make sense for me to load the same envs in the job that does the migration.

meysam81:
@Demonsthere 
You're right, it should be present for initJob as well. :+1:

helm/charts/kratos/templates/job-migration.yaml
```diff
@@ -70,6 +70,11 @@ spec:
           {{- if $migrationExtraEnv }}
             {{- toYaml $migrationExtraEnv | nindent 10 }}
           {{- end }}
+        {{- if .Values.deployment.environmentSecretsName }}
```
Demonsthere, line 73: i think the indentation here is a bit too much 😉

Demonsthere (approved):

PR #685, opened 2024-05-17
Allow self managed rules without maester disabled in PR #669

edeustua:
Since PR #669, self-managed access rules are impossible to deploy. This commit fixes that by allowing the deployment controller to load the rules config map when maester is disabled.

In PR #669, logic was added to the deployment controller such that the access rules configmap would not be loaded when `.Values.managedAccessRules` was false. This breaks down self-templated access rules when maester is not desired.

## Checklist

- [X] I have referenced an issue containing the design document if my change introduces a new feature.

## Further comments

There might be other ways to address this. Let me know if you have other suggestions. In PR #669, a suggestion was made restrict the logic only when maester's sideloader mode was enabled @cbrendanprice . This might be a solution as well.

### DIFF
--- a/helm/charts/oathkeeper/templates/deployment-controller.yaml
+++ b/helm/charts/oathkeeper/templates/deployment-controller.yaml
@@ -62,7 +62,7 @@ spec:
             name: {{ include "oathkeeper.fullname" . }}-config
             {{- end }}
         - name: {{ include "oathkeeper.name" . }}-rules-volume
-          {{- if .Values.oathkeeper.managedAccessRules }}
+          {{- if or .Values.oathkeeper.managedAccessRules ( not .Values.maester.enabled ) }}
           configMap:
             name: {{ include "oathkeeper.fullname" . }}-rules
           {{- else }}
@@ -76,7 +76,7 @@ spec:
       serviceAccountName: {{ include "oathkeeper.serviceAccountName" . }}
       automountServiceAccountToken: {{ .Values.deployment.automountServiceAccountToken }}
       initContainers:
-      {{- if (not .Values.oathkeeper.managedAccessRules) }}
+      {{- if ( not ( or .Values.oathkeeper.managedAccessRules ( not .Values.maester.enabled ) ) ) }}
         - name: init
           image: "{{ .Values.image.initContainer.repository }}:{{ .Values.image.initContainer.tag }}"
           volumeMounts:


Demonsthere:
Could you please update the `hacks/values/oathkeeper.yaml` to trigger this case on our CI? I would like to expand the tests to cover a wider array of cases

edeustua:
> Could you please update the `hacks/values/oathkeeper.yaml` to trigger this case on our CI? I would like to expand the tests to cover a wider array of cases

Just to further comment on the issue. We use oathkeeper as a subchart of our own chart, and therefore we build the access rules config map before deploying oathkeeper. We do this because we don't want to use maester at this point.

Disabling `managedAccessRules` in `hacks/values/oathkeeper.yaml` would prevent the access rules defined in it from deploying properly. Maybe you would like me to add a different test? Let me know.

Demonsthere:
My end goal would be to refactor the current setup into `hacks/values/$component/values-testcase.yaml` and have the CI iterate over all cases in the directory. This is moderately bigger refactor that needs to be done in a separate issue/pr. In this case I think we can emulate the use-case by adding a oathkeeper-rules manifest [here](https://github.com/ory/k8s/tree/master/hacks/manifests), so it will act as a pre-existing cm which has to be consumed by the chart

hacks/manifests/oathkeeper-external-accessrules.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  # The name comes from oathkeeper's subchart.
+  name: '{{ include "oathkeeper.fullname" . }}-rules'
```
Demonsthere, line 5: this is a static file, not a template, functions are not supported 😉

Demonsthere: No worries :) My idea for the CI refactor would be rather straightforward.
Current structure
```
hacks/
  values/
    oathkeeper.yaml
    keto.yaml
    ...
```

New structure
```
hacks/
  values/
    oathkeeper/
      case1.yaml
      case2.yaml
    keto/
      foo.yaml
```

This would allow us to hold multiple config options in the CI and tests all of them automatically :) 

In the quote you send i meant to deploy the cm as a static object, with a predefined `foo` name, and then feed that name as a variable in the values

PR #687, opened 2024-05-23
fix: disable maester by default

### LINKED ISSUE #512: Configmap Names for AccessRules are misaligned between oathkeeper and oathkeeper-maester
### Preflight checklist

- [X] I could not find a solution in the existing issues, docs, nor discussions.
- [X] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [X] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] This issue affects my [Ory Cloud](https://www.ory.sh/) project.
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://ory.us10.list-manage.com/subscribe?u=ffb1a878e4ec6c0ed312a3480&id=f605a41b53).

### Describe the bug

the configmapnames in oathkeeper and oathkeeper maester are not aligned and there seems to be no option to make oathkeeper actually use the config map generated by oathkeeper-maester

### Reproducing the bug

1. run helm template ory/oathkeeper
2. check the value for the --rulesConfigmapName option in the command of the oathkeeper-maester deployment
    and compare it with the configmap name in the volume section of the oathkeeper deployment

### Relevant log output

_No response_

### Relevant configuration

_No response_

### Version

0.25.3

### On which operating system are you observing this issue?

Linux

### In which environment are you deploying?

Kubernetes

kenankule:
This patch changes the default value `maester.enabled` to `false`. If the user would like to manage the rules with the maester controller, `maester.enabled` should be set to `true` and `managedAccessRules` should be set to `false`

Fixes #512
## Checklist

## Further comments

Tested with the following configurations:

1- The user would like to manage rules manually. In this configuration the rules configmap and the initContainer to fix the permissions are not added to the deployment. rules volume is an emptydir:
```
oathkeeper:
  managedAccessRules: false
maester:
  enabled: false
```

2- The rules are managed by the helm chart `oathkeeper.accessRules` value. Rules configmap is created but the initContainer is not added to the deployment.
```
oathkeeper:
  managedAccessRules: true
maester:
  enabled: false
```

3- The rules are managed by the maester. Rules configmap is created by the maester chart and the volume name in the oathkeeper deployment is aligned with the maester chart. initContainer is not added to the deployment.
```
oathkeeper:
  managedAccessRules: false
maester:
  enabled: true
```

4- Invalid configuration so the helm chart rendering fails with an error message.
```
oathkeeper:
  managedAccessRules: true
maester:
  enabled: true
```

I believe the first configuration (where the chart creates an emptydir volume) should be replaced with a configuration where the chart gets an existing configmap name and uses that instead.

### DIFF
--- a/hacks/values/oathkeeper.yaml
+++ b/hacks/values/oathkeeper.yaml
@@ -68,6 +68,7 @@ deployment:
       maxSurge: 25%
       maxUnavailable: 25%
 oathkeeper:
+  managedAccessRules: true
   accessRules: |
     [
       {
--- a/helm/charts/oathkeeper/templates/_helpers.tpl
+++ b/helm/charts/oathkeeper/templates/_helpers.tpl
@@ -24,6 +24,21 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 {{- end -}}
 
+
+{{/*
+Create a config map name for rules.
+If maester is enabled, use the child chart named template to get the value.
+*/}}
+{{- define "oathkeeper.rulesConfigMapName" -}}
+{{- if .Values.maester.enabled -}}
+{{- $childChart := (dict "Name" "oathkeeper-maester") -}}
+{{- include "oathkeeper-maester.getCM" (dict "Values" (index .Values "oathkeeper-maester") "Release" $.Release "Chart" $childChart) }}
+{{- else -}}
+{{ include "oathkeeper.fullname" . }}-rules
+{{- end -}}
+{{- end -}}
+
+
 {{/*
 Create a secret name which can be overridden.
 */}}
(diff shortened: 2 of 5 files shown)


eroznik:
Any info on this one? Whats the hold up? :thinking: 
AFAIK, without this we're not able to use the chart as is - there is no way to make the Oathkeeper chart not create "empty rules" but still using the config map created by Maester ..

kenankule:
I've created this fix only for a PoC and we're not using the chart or the product anymore. I'd be happy if the maintainers take the ownership of this changeset.

Demonsthere:
Hi there!
The issue with this change is that it changes the default behaviour for people who do not use the maester controller, which we consider an addon. If we could provide the same function without changing the default behaviour (as still, most people use oathkeeper standalone, without the controller) i would be more than happy to merge this :)

kenankule:
I think that (change in the default behavior) can be avoided if I set 
```
managedAccessRules: true
maester:
  enabled: false
```
in the default values of the chart. Should I go ahead and change the defaults so that the change is somewhat backwards compatible? The only case it is not is when both of these values are set to `true` which I believe should not happen. There is even a [comment](https://github.com/ory/k8s/blob/master/helm/charts/oathkeeper/values.yaml#L203) in the values file about this.

Demonsthere:
@kenankule yes please :) I think it even would make sense to add a check for this case (both are set to true) and fail the installation with an error message when that happens.

kenankule:
I've changed the defaults, rebased the branch and pushed. Also edited the changeset description.

The validation you've requested is already implemented : https://github.com/ory/k8s/blob/6788f474786e7743a3867bdd769c99a425a56e68/helm/charts/oathkeeper/templates/configmap-rules.yaml#L3

Demonsthere (approved):

eroznik:
@Demonsthere any ETA for the next  release? :))

PR #693, opened 2024-07-04
chore: upgrade kratos image to v1.2.0

meysam81:
## Checklist

### DIFF
--- a/helm/charts/kratos/Chart.yaml
+++ b/helm/charts/kratos/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v2
-appVersion: "v1.1.0"
+appVersion: "v1.2.0"
 description: A ORY Kratos Helm chart for Kubernetes
 name: kratos
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-kratos.svg
-version: 0.46.0
+version: 0.45.0
 type: application
 dependencies:
   - name: ory-commons
--- a/helm/charts/kratos/Chart.yaml
+++ b/helm/charts/kratos/Chart.yaml
@@ -3,7 +3,7 @@ appVersion: "v1.1.0"
 description: A ORY Kratos Helm chart for Kubernetes
 name: kratos
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-kratos.svg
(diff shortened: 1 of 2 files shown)


helm/charts/kratos/Chart.yaml
```diff
@@ -3,7 +3,7 @@ appVersion: "v1.1.0"
 description: A ORY Kratos Helm chart for Kubernetes
 name: kratos
 icon: https://raw.githubusercontent.com/ory/docs/master/docs/static/img/logo-kratos.svg
-version: 0.45.0
+version: 0.46.0
```
Demonsthere, line 6: this is the version of the chart, that is autogenerated on release, please don't touch ;)

Demonsthere: You can update `appVersion: "v1.1.0"` to reflect the new version

Demonsthere (changes requested):
Please correct the versions in chart.yaml and run `make helm-docs` :)

helm/charts/kratos/Chart.yaml
meysam81: @Demonsthere 

While I'm gonna go ahead and apply your suggestion, I think this is a good moment for me to ask, why would you upgrade all the charts at the same version, even if others have not had change?

Wouldn't you say it is more appropriate to have per-chart Chart-version? based don't the changes of specific chart.

Demonsthere: I agree it can be confusing, it's a pattern that we created when all of the charts were in a active development model. We are unsure if we want to keep the charts in this repo, or move them to the corresponding application repos, so for now we just release them as a bunch

meysam81: If it helps in any way, I think the code structure of Helm charts is better on its own repository. That's usually more maintainable in terms of the frequency of code changes for application code as compared to infra/platform/IaC code.

Having one helm chart per repository is not feasible IMO, but I believe the current repository can be structured in a better way so that Chart versions are not bumped at the same time, each having their own pace.

It can result into confusion otherwise when one of the charts need a bump and others don't, and yet all get a bump anyway!

Demonsthere (approved):

PR #706, opened 2024-10-01
feat(hydra): option to create separate admin and public deploys

terev:
Add an option to Hydra chart `separateAdminAndPublicDeploys`, when enabled creates separate deployment objects for the admin and public components. In addition, if auto-scaling is enabled, separate HPA objects will be created.

I was unable to find any related issues, and don't have a design document.

## Checklist

## Further comments

The motivation for this change was to enable creating an Istio `RequestAuthentication` policy which applies to only the admin component (since it doesnt have its own auth). In addition, this change makes it possible to scale, configure, and route the admin and public components separately.


Demonsthere:
Hi there!
This is an interesting feature request, I can see the reasoning behind it, and it possibly would make sense to create such a distributed system for other applications as well 🤔.
However, I think it might be a better idea to make it more verbose, and follow the same path as https://github.com/grafana/helm-charts/tree/main/charts and create a separate hydra-distributed chart, as opposed to keeping it bundled in a single package, wdyt?

terev:
@Demonsthere Are there any concerns with the additional complexity required to update two charts for Hydra if we fork this one into a new chart? My concern is if there's multiple different charts it gets complicated to keep them all maintained. Maybe we could modularize this chart such that portions can be reused in the distributed version?

terev:
@Demonsthere If it wasn't clear, I'm totally open to doing this. Just looking for guidance to do it in a maintainable way.

terev:
@Demonsthere I essentially copied the hydra chart (including my changes) to a new `hydra-distributed` chart. Then I removed the `separateAdminAndPublicDeploys` option that I added.

Is this in line with what you were thinking?

terev:
@Demonsthere Is the new approach in line with what you were thinking?

PR #725, opened 2024-12-17
feat: allow user to scale more precisely with ContainerResource in hpa

noony:

Add possibility to define extra containers metrics in hpa to scale independently and more precisely when multiple containers are declared.

## Checklist

### DIFF
--- a/helm/charts/oathkeeper/templates/hpa-controller.yaml
+++ b/helm/charts/oathkeeper/templates/hpa-controller.yaml
@@ -1,4 +1,5 @@
 {{- if .Values.deployment.autoscaling.enabled }}
+{{- $kubeVersion := .Capabilities.KubeVersion.GitVersion -}}
 apiVersion: autoscaling/v2
 kind: HorizontalPodAutoscaler
 metadata:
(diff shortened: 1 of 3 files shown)


helm/charts/oathkeeper/templates/hpa-controller.yaml
```diff
@@ -20,17 +21,34 @@ spec:
   maxReplicas: {{ .Values.deployment.autoscaling.maxReplicas }}
   metrics:
   {{- with .Values.deployment.autoscaling.targetMemory }}
+  {{- if and (and (not (empty $kubeVersion)) (semverCompare ">=1.30-0" $kubeVersion) (not (empty $.Values.deployment.autoscaling.extraMetrics))) }}
```
noony, line 24: to avoid breaking change for people already using extraContainers, we switch to containerResource only when it's available in kubernetes > 1.30 and Values.deployment.autoscaling.extraMetrics is defined.

David-Wobrock:
> I like the idea, but not 100% sold on the implementation 🤔 I was wondering if we should not proceed on a cleaner path:
> 
>     * define a new template for hpa, that is for k8s 1.30+ where we use only ContainerResources, as from my pov we should scale on the base of the core ory apps, and not sidecars
> 
>     * then at 1.3x deprecate the older file and leave only ContainerResource scaling
> 
>     * also, regardless of which path we choose, i would prefer to have it across all charts, and not only oathkeeper ;)
> 
> 
> wdyt?

Sounds good!
I'm picking up the topic, see here a new PR: https://github.com/ory/k8s/pull/730 :)

David-Wobrock:
I think we can close this PR, as https://github.com/ory/k8s/pull/730 was merged and released in [v0.51.0](https://github.com/ory/k8s/releases/tag/v0.51.0).

Issue #753, opened 2025-03-08
Add `extraObjects` to the helm chart


khusseini:
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [x] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [x] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

## Summary

Introduce an extraObjects section in the Helm values to allow adding arbitrary Kubernetes resources to the Oathkeeper chart.

### Motivation

- Automate initial JWT file rotation after Oathkeeper installation 
- Eliminate manual intervention in the setup process
- Enable seamless integration of custom resources without creating separate Helm charts

### Benefits

- Streamlines post-installation tasks
- Increases flexibility for custom deployments
- Simplifies management of Oathkeeper-related resources

### Describe your ideal solution

Add `extraObjects` key in values.yaml and process them and add the following file (`extra-manifests.yaml`) to the `templates` directory:

```
{{ range .Values.extraObjects }}
---
{{ tpl (toYaml .) $ }}
{{ end }}
```

### Workarounds or alternatives

Currently the workaround is to create a new Helm Chart which depends on oathkeeper in order to add the above mentioned template. 

### Version

Chart version 0.52.0

khusseini:
I took the liberty and added a PR for this issue

PR #754, opened 2025-03-08
feat: added extraObjects to Oathkeeper chart

khusseini:

Implements #753 

## Checklist


Demonsthere (changes requested):
Hello there! 
I don't have any issues with the feature, but I see how we could reuse it easier in other charts. Please ass the extra-manifest as a template to the https://github.com/ory/k8s/tree/master/helm/charts/ory-commons/templates library and we can easily include it from there. This will make reusing this much simpler :)

khusseini:
@Demonsthere I moved the template to ory-commons helm chart templates folder. Sorry it took me a while to do this.

Demonsthere (changes requested):
a few things here, as this wont work this way. The ory-commons is a library chart, and doesnt have Values or manifests, but only functions like https://github.com/ory/k8s/blob/master/helm/charts/ory-commons/templates/_helpers.tpl#L4 
You need to refactor the extra manifests into a template function, then add the library as a dependency to oathkeeper as in https://github.com/ory/k8s/blob/master/helm/charts/keto/Chart.yaml#L29 
then you have access to the functions inside which can be called similar to https://github.com/ory/k8s/blob/master/helm/charts/keto/templates/deployment.yaml#L97

So in the end, you should have in oathkeeper the extra-manifests.yaml file with a structure similar to 

```yaml
{{- with .Values.extraObjects }}
{{ include "ory.generateExtraObjects" . }}
{{- end }}
```

Demonsthere:
Hey there @khusseini no worries ;) I have written more details on how i think this change should be implemented. If you have any questions, please ping me 😉

yo-ga:
I’d like to ask if there has been any progress on this. 
Also, will it support all charts? We need this for handling Hydra in the near future.

Issue #770, opened 2025-05-12
Secret annotations for helm hook on kratos/hydra/oathkeeper can't be deactivated


maximumG:
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [ ] I have joined the [Ory Community Slack](https://slack.ory.sh).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.sh/l/sign-up-newsletter).

### Describe the bug

We are using the official kratos helm chart to deploy kratos in our environment using helm chart version `v0.53.0`. This chart is deployed with helm `v3.16.4`.

This version adds some [helm hook annotations on Secrets](https://github.com/ory/k8s/blob/63cea7f05e2955c78457d2172e966cfe5dd617b4/helm/charts/kratos/values.yaml#L86-L92) in order to generate a new Secret upon each helm deployment of the chart and randomize some secrets.

We do not want these annotations to be set on Secret for lifecycle purpose while still want to the chart to create the Secret (The Secret value are coming from a external secret store).

We tried to reset this annotation by overriding the following value with `null`.

```yaml
secret:
  secretAnnotations:
    helm.sh/hook-weight: null
    helm.sh/hook: null
    helm.sh/hook-delete-policy: null
    helm.sh/resource-policy: null
```

However because of a bug in helm `v3.16.4`this annotation are not dropped when helm merges the keys using a subchart : https://github.com/helm/helm/pull/12879

Would it be possible to add this annotations as an opt-in/opt-out fashion ?

### Reproducing the bug

1. Use helm `v3.16.4`
2. Create a new YAML file to override the Secret annotations and use a subchart

* `Chart.yaml` file
```yaml
apiVersion: v2
name: kratos
version: 0.1.0
appVersion: "1.0"
type: application
keywords:
  - helm
dependencies:
  - name: kratos
    version: 0.53.0
    repository: https://k8s.ory.sh/helm/charts
```

* `values.yaml` file
```yaml
secret:
  secretAnnotations:
    helm.sh/hook-weight: null
    helm.sh/hook: null
    helm.sh/hook-delete-policy: null
    helm.sh/resource-policy: null
```

[...]

PR #771, opened 2025-05-13
fix: helm hooks can be opt-out on ory components

maximumG:
Helm hooks can be easily opt-out from any ory component's charts when creating Kubernetes Secret.

#770 

## Checklist

## Further comments

N/A

### DIFF
--- a/helm/charts/hydra/templates/_helpers.tpl
+++ b/helm/charts/hydra/templates/_helpers.tpl
@@ -83,6 +83,18 @@ Generate the name of the secret resource containing secrets
 {{- end -}}
 
 {{/*
+Generate the secrets.annotations value
+*/}}
+{{- define "hydra.secrets.annotations" -}}
+  {{- $annotations := dict -}}
+  {{- if .Values.secret.enableDefaultAnnotations }}
+    {{- $annotations = .Values.secret.secretAnnotations -}}
+  {{- end -}}
+  {{- $annotations = merge $annotations .Values.secret.extraAnnotations -}}
+  {{- toYaml $annotations }}
+{{- end -}}
+
+{{/*
 Generate the secrets.system value
 */}}
 {{- define "hydra.secrets.system" -}}
--- a/helm/charts/hydra/README.md
+++ b/helm/charts/hydra/README.md
@@ -174,7 +174,9 @@ A Helm chart for deploying ORY Hydra in Kubernetes
 | pdb.spec.minAvailable | string | `""` |  |
 | priorityClassName | string | `""` | Pod priority https://kubernetes.io/docs/concepts/configuration/pod-priority-preemption/ |
 | replicaCount | int | `1` | Number of ORY Hydra members |
+| secret.enableDefaultAnnotations | bool | `true` | enableDefaultAnnotations set to `true` will add default annotations to the secret. As such the Secret will be managed by helm hooks. |
 | secret.enabled | bool | `true` | switch to false to prevent creating the secret |
+| secret.extraAnnotations | object | `{}` | extraAnnotations to be added to secret. |
 | secret.hashSumEnabled | bool | `true` | switch to false to prevent checksum annotations being maintained and propogated to the pods |
 | secret.nameOverride | string | `""` | Provide custom name of existing secret, or custom name of secret to be created |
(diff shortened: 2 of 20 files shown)


Demonsthere:
Hello there!
I see the point of the PR, that by overriding the annotations, you can remove the hook annotations and break your installation. Now sure though if i like the solution 🤔. 
I wonder if it wouldn't be cleaner to leave the secretAnnotations as is, and add an extra field like `secret.extraAnnotations` which would be empty, and merged o the default annotations. This would leave explicit default annotations that if required can be overwritten, while still allowing users to add extra annotations. Wdyt?

maximumG:
Hey thanks for your quick feedback on this, really appreciated 😃 .

> I wonder if it wouldn't be cleaner to leave the secretAnnotations as is, and add an extra field like secret.extraAnnotations which would be empty, and merged o the default annotations. This would leave explicit default annotations that if required can be overwritten, while still allowing users to add extra annotations

From a design perspective I couldn't agree more. However, due to the helm bug https://github.com/helm/helm/pull/12879, I think we'll end up in the same situation, where we can't remove helm hooks annotations when using ory charts as a **subchart**. Let me explain.

The **default** values provided by the chart would look like this
```yaml
secret:
  secretAnnotations: # default to use helm hooks
    helm.sh/hook-weight: "0"
    helm.sh/hook: "pre-install, pre-upgrade"
    helm.sh/hook-delete-policy: "before-hook-creation"
    helm.sh/resource-policy: "keep"
  extraAnnotations: {}
```

If I use ory chart as a subchart and the following values, merging the keys won't work as expected.  We end up in a situation where the helm hooks annotations will still be present in the rendered manifest.

```yaml
secret:
  secretAnnotations: {} # this will be merged with default values above and we will have helm hooks annotations
  extraAnnotations:
    my-custom-annotation: my-custom-values # I will also have my custom-annotations, that's cool 😃 
```

That's why we would need an explicit opt-out to remove explicitly helm hooks annotations, without hitting the helm merging keys bug I outlined in my issue.

Demonsthere:
Right, i see the use-case now 🤔 
The only thing i don't like in your PR is the implicit definition of annotations, which is hidden and cumbersome to find/update: 
```
{{- $hooks := dict "helm.sh/hook-weight" "0" "helm.sh/hook" "pre-install, pre-upgrade" "helm.sh/hook-delete-policy" "before-hook-creation" "helm.sh/resource-policy" "keep" -}}
```

I think we could combine the two approaches and have the values in the following format:
```yaml
secret:
  secretAnnotations: # default to use helm hooks
    helm.sh/hook-weight: "0"
    helm.sh/hook: "pre-install, pre-upgrade"
    helm.sh/hook-delete-policy: "before-hook-creation"
    helm.sh/resource-policy: "keep"
  extraAnnotations: {}
  enableDefaultAnnotations: true
 ```
 
 this would allow you to do the following:
```yaml
keto:
  secret:
    enableDefaultAnnotations: false
    extraAnnotations:
      my-custom-annotation: my-custom-values # I will also have my custom-annotations, that's cool 😃 
```

but keep the definition of the default hooks in a explicit way so that `helm show values` will still show them, as well as the readme :)

maximumG:
> I think we could combine the two approaches and have the values in the following format

Indeed this combination is a really good trade-off. I'll push the modification next week

maximumG:
@Demonsthere Here is the implementation we talked about. Hope it fits your vision as well 🤞

helm/charts/hydra/templates/_helpers.tpl
```diff
@@ -82,6 +82,18 @@ Generate the name of the secret resource containing secrets
...
+{{/*
+Generate the secrets.annotations value
+*/}}
+{{- define "hydra.secrets.annotations" -}}
+  {{- $annotations := dict -}}
+  {{- if .Values.secret.enableDefaultAnnotations }}
+    {{- $annotations = .Values.secret.secretAnnotations -}}
+  {{- end -}}
+  {{- $annotations = merge $annotations .Values.secret.extraAnnotations -}}
+  {{- toYaml $annotations }}
```
Demonsthere, line 94: i think we could just take the extraAnnotations and do the conditional merge to it, but it's programmatically the same :)

Demonsthere (approved):
besides a small code nitpick, lgtm :D I think we can merge and release as is

PR #804, opened 2025-09-21
feat: add relabeling and metricRelabeling in serviceMonitor for all helm charts

fedeabih:
# Add relabeling and metricRelabeling for all helm charts

This pull request updates all Helm charts to support feat-relabeling and metric relabeling.
It introduces new configuration options that allow users to customize relabeling rules for Prometheus metrics
across all charts in a consistent way.

Fix: #805

## Checklist

### DIFF
--- a/hacks/values/hydra-maester/default.yaml
+++ b/hacks/values/hydra-maester/default.yaml
@@ -38,3 +38,13 @@ serviceMonitor:
     release: "prometheus"
   tlsConfig:
     insecureSkipVerify: true
+  relabelings:
+  - source_labels: [__address__]
+    regex: '([^:]+):.*'
+    target_label: instance
+    replacement: '$1'
+  metricRelabelings:
+  - source_labels: [__address__]
+    regex: '([^:]+):.*'
+    target_label: instance
+    replacement: '$1'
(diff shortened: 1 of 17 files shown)


Demonsthere (changes requested):
Hey there! 
Some housekeeping before we can merge this :) 
1. Please create at least 1 new test values overrides file (in `hacks/values/app`) that will have some relabeling set, so that the CI can generate a serviceMonitor with the fields set
2. Please run `make helm-docs` to update chart readme with the new values

fedeabih:
> Hey there! Some housekeeping before we can merge this :)
> 
> 1. Please create at least 1 new test values overrides file (in `hacks/values/app`) that will have some relabeling set, so that the CI can generate a serviceMonitor with the fields set
> 2. Please run `make helm-docs` to update chart readme with the new values

Hi! I followed your feedback, but instead of creating a new values override file, I added the relabeling example directly to the default values. Since the change is small and doesn’t introduce much complexity, I thought it might be simpler to keep it in the default file rather than maintaining an extra override file.

Would you be okay with that approach, or do you prefer that I still create a separate file under hacks/values/app?

Regards!

Demonsthere:
I am ok with setting that in the default values :) I see that the CI fails when installing postgres 🤔 will check that when i can, as possibly the overrides needs to be updated for the db

PR #879, opened 2026-06-17
feat: add per-job extra volumes to hydra migration job

erichwoo-scale:
The hydra migration `Job` currently has no way to mount volumes independently of the long-running `deployment`. The `volumes` / `volumeMounts` blocks in `templates/job-migration.yaml` are hardcoded to `.Values.deployment.extraVolumes` and `.Values.deployment.extraVolumeMounts`, so the only way to attach a volume to the short-lived migration pod is to also attach it to the server deployment.

This is inconsistent with the rest of that same template, where `job.nodeSelector`, `job.extraEnv`, and `job.resources` already override their `deployment.*` / `automigration.*` counterparts via a `ternary` fallback. This PR closes that gap by adding `job.extraVolumes` and `job.extraVolumeMounts`, following the identical fallback pattern:

```yaml
job:
  extraVolumes:
    - name: my-volume
      secret:
        secretName: my-secret
  extraVolumeMounts:
    - name: my-volume
      mountPath: /etc/secrets/my-secret
      readOnly: true
```

**Backward compatible:** both default to `[]`. When unset, the job continues to fall back to `deployment.extraVolumes` / `deployment.extraVolumeMounts`, so existing releases render byte-for-byte identically. The change also converts the two blocks from `if` to `with` and fixes a mis-indented `{{- end }}` in the existing `volumeMounts` block.

No linked issue. This is a small consistency enhancement rather than a new feature: it extends the pre-existing `job.* → deployment.*` override pattern in `templates/job-migration.yaml` (already present for `nodeSelector`, `extraEnv`, and `resources`) to volumes. Happy to open a discussion first if maintainers prefer.

## Checklist

## Further comments

**Alternative considered:** keeping the volumes coupled to `deployment.extraVolumes`. Rejected because it forces unrelated volumes (e.g. a migration-only credential mount) onto the always-running server pod, and because every sibling setting in the same template already supports a per-job override — this just makes volumes consistent.

The `hacks/values/hydra/default.yaml` change adds an `emptyDir`-backed `job.extraVolumes` + `job.extraVolumeMounts` so the existing render/kubeconform CI exercises the new code path.

## Summary by CodeRabbit

* **New Features**
  * Added job-level `extraVolumes` and `extraVolumeMounts` configuration options for Hydra Helm chart. These settings take precedence over deployment-level volume configurations when specified.

* **Documentation**
  * Updated Helm chart documentation to reflect new job-specific volume configuration options.

### DIFF
--- a/hacks/values/hydra/default.yaml
+++ b/hacks/values/hydra/default.yaml
@@ -154,6 +154,12 @@ job:
       image: "alpine:latest"
       command: ["/bin/sh"]
       args: ["-c", "sleep 10"]
+  extraVolumes:
+    - name: job-extra
+      emptyDir: {}
+  extraVolumeMounts:
+    - name: job-extra
+      mountPath: /job-extra
   shareProcessNamespace: true
   podMetadata:
     labels:
--- a/helm/charts/hydra/README.md
+++ b/helm/charts/hydra/README.md
@@ -161,6 +161,8 @@ A Helm chart for deploying ORY Hydra in Kubernetes
 | job.extraContainers | string | `""` | If you want to add extra sidecar containers. |
 | job.extraEnv | list | `[]` | Array of extra envs to be passed to the job. This takes precedence over deployment variables. Kubernetes format is expected. Value is processed with Helm `tpl` - name: FOO   value: BAR |
 | job.extraInitContainers | string | `""` | If you want to add extra init containers. extraInitContainers: |  - name: ...    image: ... |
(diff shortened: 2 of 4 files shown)


Demonsthere (approved):

Demonsthere:
Hey there! 
I assume this wasn't added before, because simply no use case was reported, but i agree that it makes the job more consistent with the rest :)

erichwoo-scale:
Thanks! Will keep my eye out for the next release :)

erichwoo-scale:
@Demonsthere curious when should I expect to see v0.62.2 release cut? Thanks!

PR #883, opened 2026-07-01
fix: add namespace to oathkeeper poddisruptionbudget helm template

naohiroo:
## Related Issue or Design Document

https://github.com/ory/k8s/issues/882

## Checklist

- [x] I have read the [contributing guidelines](../blob/master/CONTRIBUTING.md) and signed the CLA.
- [x] I have referenced an issue containing the design document if my change introduces a new feature.
- [x] I have read the [security policy](../security/policy).
- [x] I confirm that this pull request does not address a security vulnerability. 
      If this pull request addresses a security vulnerability, 
      I confirm that I got approval (please contact [security@ory.com](mailto:security@ory.com)) from the maintainers to push the changes.
- [ ] I have added tests that prove my fix is effective or that my feature works.
- [ ] I have added the necessary documentation within the code base (if appropriate).

## Summary by CodeRabbit

* **Bug Fixes**
  * Pod disruption budgets now include the release namespace when available, ensuring they’re created in the correct location.

Issue #882, opened 2026-07-01
oathkeeper PodDisruptionBudget helm template is missing namespace metadata


naohiroo:
### Preflight checklist

- [x] I could not find a solution in the existing issues, docs, nor discussions.
- [x] I agree to follow this project's [Code of Conduct](https://github.com/ory/k8s/blob/master/CODE_OF_CONDUCT.md).
- [x] I have read and am following this repository's [Contribution Guidelines](https://github.com/ory/k8s/blob/master/CONTRIBUTING.md).
- [x] I have joined the [Ory Community Slack](https://slack.ory.com).
- [ ] I am signed up to the [Ory Security Patch Newsletter](https://www.ory.com/l/sign-up-newsletter).

### Describe the bug

- oathkeeper `PodDisruptionBudget` helm template is missing namespace metadata.
    - https://github.com/ory/k8s/blob/v0.62.1/helm/charts/oathkeeper/templates/pdb.yaml#L5-L6

### Reproducing the bug

1. Run the following command:
   ```bash
   helm template my-release ory/oathkeeper \
     --version 0.62.1 \
     --namespace my-namespace \
     --set pdb.enabled=true \
     --show-only templates/pdb.yaml
   ```
2. Check the output:
      ```yaml
      # Source: oathkeeper/templates/pdb.yaml
      apiVersion: policy/v1
      kind: PodDisruptionBudget
      metadata:
        name: my-release-oathkeeper
      spec:
        selector:
          matchLabels:
            app.kubernetes.io/name: oathkeeper
            app.kubernetes.io/instance: my-release
      ```
    

### Relevant log output

```shell

```

### Relevant configuration

```yml

```

### Version

v0.62.1

### On which operating system are you observing this issue?

None

### In which environment are you deploying?

None

### Additional Context

It should be the following:
```yaml
# Source: oathkeeper/templates/pdb.yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: my-release-oathkeeper
  namespace: my-namespace
spec:
  selector:
    matchLabels:
      app.kubernetes.io/name: oathkeeper
      app.kubernetes.io/instance: my-release
```

PR #888, opened 2026-07-03
fix: add namespace to remaining poddisruptionbudget helm templates

somaz94:
The PodDisruptionBudget templates for kratos, hydra-maester, and
oathkeeper-maester don't set `metadata.namespace`, so the PDB doesn't
inherit the release namespace. This completes the same fix already
merged for hydra (#885) and in progress for oathkeeper (#883); keto
already has it.

Same guard as #885:

```
  {{- if .Release.Namespace }}
  namespace: {{ .Release.Namespace }}
  {{- end }}
```

Validation:
- `helm template` on all three charts with `pdb.enabled=true` renders
  `namespace: <release-namespace>` on the PDB.
- default render is unchanged; the only delta vs master is the added
  namespace line.

related: #882, #884

## Summary by CodeRabbit

* **Bug Fixes**
  * PodDisruptionBudget resources now include the release namespace when one is provided, improving chart rendering consistency across deployments.
  * Applied the same namespace handling across all affected charts to reduce surprises during installation or upgrade.

### DIFF
--- a/helm/charts/hydra-maester/templates/pdb.yaml
+++ b/helm/charts/hydra-maester/templates/pdb.yaml
@@ -4,6 +4,9 @@ apiVersion: policy/v1
 kind: PodDisruptionBudget
 metadata:
   name: {{ include "hydra-maester.fullname" . }}
+  {{- if .Release.Namespace }}
+  namespace: {{ .Release.Namespace }}
+  {{- end }}
 spec:
   selector:
     matchLabels:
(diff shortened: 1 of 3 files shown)


somaz94:
Hi @Demonsthere — friendly ping on this one.

All checks are green and there are no conflicts. This is the follow-up to #885 (hydra), covering the remaining charts that still template a PodDisruptionBudget without a namespace: hydra-maester, kratos and oathkeeper-maester. It does not overlap with #885 or with the open #883.

Happy to rebase or split it up if that would help. Thanks!

Demonsthere (approved):

Demonsthere:
Hi there!
Sorry for the delay on this 😞

